Issue #22, opened 2020-03-20
Improve rescan triggering


charvp:
Should keep refreshing until server has started rescan and only then fall back to the automatic refresh of the rescan action.

PR #44, opened 2020-03-21
Add object to vuex when fetched

robbevp:
# TODO
* [x] Figure out how to delete outdated object on refresh


robbevp:
Same discussion as on gitlab:
We still need to figure out how to remove outdated objects on refresh. 
I'm still in favor of adding a 'last loaded' property to all objects. This might also come in handy if we ever decide to load some content between refreshing (for example, before opening an edit view)

charvp:
Another idea I had: We can just add the objects on the fly when loading, but still do a big replace in the end. One thing that would need extra attention in that case is that, while loading, edits made to items that were already loaded would probably be lost. They would still be correct on the server, but the user would see the old values again. (Note that this can already happen, in fact, I've been bitten by that one a few times.)

One solution to that extra snag is to disallow edits while loading, but I'm not a huge fan of that.

robbevp:
> Another idea I had: We can just add the objects on the fly when loading, but still do a big replace in the end. One thing that would need extra attention in that case is that, while loading, edits made to items that were already loaded would probably be lost. They would still be correct on the server, but the user would see the old values again. (Note that this can already happen, in fact, I've been bitten by that one a few times.)
> 
> One solution to that extra snag is to disallow edits while loading, but I'm not a huge fan of that.

But then doesn't that create more problems than it solves?
One benefit of knowing when an item was last loaded, is that we can disallow edits while loading for the shortest possible time. We can already enable this again for those items that have already loaded.

Also: If we have can load the data faster (accentor/api#22 and accentor/api#23) and cache the whole store #48, I'm not sure if this still adds a lot of benefit.

charvp:
> But then doesn't that create more problems than it solves?
> One benefit of knowing when an item was last loaded, is that we can disallow edits while loading for the shortest possible time. We can already enable this again for those items that have already loaded.

I was going to write a spiel about how I still prefer the big update at the end, but your "disallow edits while loading for the shortest possible time" argument (sort of) convinced. I say sort of, because the actual benefit is that while loading, we won't even replace the edited content with outdated data. Since either it was loaded already, and then it won't be replaced again, or the server already has the edit, and the data will be replaced with the correct edited data. Only if the edit happens between the time the request was started for the page the edited item is in and the time the request finishes, the item will still be overwritten with old data, but that seems like a rare enough case to me.

> Also: If we have can load the data faster ([accentor/api#22](https://github.com/accentor/api/pull/22) and [accentor/api#23](https://github.com/accentor/api/pull/23)) and cache the whole store #48, I'm not sure if this still adds a lot of benefit.

The performance improvements recently added to API aren't significant enough to load the data very quickly: It still takes a bit more than 1.5 seconds until all tracks are loaded. #48 didn't turn out to be feasible due to performance impact, so I do think this still adds a lot of benefit. (And like we discussed, getting the new content almost immediately after a rescan seems like a big benefit to me.)

charvp:
To solve the last edge-case I mentioned in my previous comment, we could use the `updated_at` property. If it is newer than the time the request was started (which is the same as the new `loaded` value, I would assume), you could just not update the item.

There are a few tables that don't have timestamps (or don't expose them), but if we fix this in the API, that seems like a good solution to me.

charvp:
This can also close #64 if you think my solution would work.

src/api/fetch.js
```diff
@@ -0,0 +1,18 @@
+import baseURL from "./base_url";
+
+export function fetchIndex(url, page, auth) {
```
charvp: Just `index` as a method name is fine. The fact that it is in the `api` module makes clear that it will do fetching. Also, rename `url` to `path` since it's not a URL, it's a path.

src/api/albums.js
```diff
@@ -1,37 +1,5 @@
 import baseURL from "./base_url";
 
-export function index(auth) {
```
charvp: I would prefer if you just re-export the generic fetch method here with the first argument set. You can use a function reference in the store to still have a generic commit action.

src/store/commit.js
```diff
@@ -0,0 +1,35 @@
+import { fetchIndex } from "../api/fetch";
+
+export function index({ commit, auth }, { url, commitAction }) {
```
charvp: If the `api` `index` is a generator, I don't think this function should still be called `index`. Something like `fetchAll` maybe?

robbevp:
> To solve the last edge-case I mentioned in my previous comment, we could use the `updated_at` property. If it is newer than the time the request was started (which is the same as the new `loaded` value, I would assume), you could just not update the item.

> This can also close #64 if you think my solution would work.

I would still disable editing until the property is re-loaded. If an obj is updated while you are editing, you lose your work. 
Moreover, there is a chance that you update an obj which was already updated by another user. (This is something for a different PR, but I would to get the latest version from the API before a user can edit - certainly for the major models)

charvp:
> I would still disable editing until the property is re-loaded. If an obj is updated while you are editing, you lose your work.
> Moreover, there is a chance that you update an obj which was already updated by another user. (This is something for a different PR, but I would to get the latest version from the API before a user can edit - certainly for the major models)

Both of those seem separate issues to me.

robbevp:
@charvp this might still be rough in some parts, but it is a working second draft using a generator

src/api/albums.js
robbevp: This is possible, but it feels weird to me. Do you still want this now that there is a generator function?

charvp: > The store shouldn't need to know what the HTTP path is for this resource.

This is actually the main reason I want this. Seperation of concerns is important to me.

> Once we convert all the common api calls into generators, most of the files in /api would only serve to redefine functions.

I think only index can be generalized into a generator. I don't see how the other functions will provide intermediate results.

charvp:
> @charvp this might still be rough in some parts, but it is a working second draft using a generator

I'll take a look when I find some time.

src/api/albums.js
robbevp: > > The store shouldn't need to know what the HTTP path is for this resource.
> 
> This is actually the main reason I want this. Seperation of concerns is important to me.

But then couldn't we also define an object that defines the path for each collection? Then we could just tell the `indexGenerator` which collection it is dealing with

> > Once we convert all the common api calls into generators, most of the files in /api would only serve to redefine functions.
> 
> I think only index can be generalized into a generator. I don't see how the other functions will provide intermediate results.

You're right, i got confused. What I meant was that we could convert most of the api calls to a more generic functions. (Since often the `path` is the only thing that is different between all the actions of the same type).

src/api/paths.js
```diff
@@ -0,0 +1,16 @@
+const paths = {
```
charvp, line 1: I still don't really like this solution. Re-exporting the functions allows us the most flexibility in the future if (for example) we need to change the implementation for one resource. It's also almost the exact same amount of code.

In a better language, this could be done with rich enums but javascript doesn't have that.

src/api/fetch.js
```diff
@@ -0,0 +1,30 @@
+import baseURL from "./base_url";
+import paths from "./paths";
+
+export async function* indexGenerator(collection, auth, page = 1) {
+  const path = paths[collection];
+  while (path) {
```
charvp, line 6: Wut?

src/api/fetch.js
```diff
@@ -0,0 +1,30 @@
...
+        (request.headers.has("x-total-pages") &&
+          request.headers.get("x-total-pages") == page) ||
+        result.length === 0
+      ) {
+        return result;
+      } else {
+        yield result;
+      }
+    } else {
+      return Promise.reject(result);
```
charvp, line 26: Returning a promise in an async function doesn't make much sense. `throw result` is what we want here I think.

src/api/fetch.js
```diff
@@ -0,0 +1,30 @@
...
+        "x-secret": auth.secret,
+        "x-device-id": auth.device_id,
+      },
+    });
+    const result = await request.json();
+    if (request.ok && result) {
+      if (
+        (request.headers.has("x-total-pages") &&
+          request.headers.get("x-total-pages") == page) ||
+        result.length === 0
```
charvp, line 19: Why is this condition still here? This shouldn't happen right? Same for the check that the header is present.

src/store/albums.js
```diff
@@ -1,34 +1,49 @@
...
-import { create, destroy, index, update, destroyEmpty } from "../api/albums";
+import { create, destroy, update, destroyEmpty } from "../api/albums";
+import { fetchAll } from "./commit";
 import { compareStrings } from "../comparators";
 
 export default {
   namespaced: true,
   state: {
     albums: {},
+    startLoading: new Date(0),
```
charvp, line 10: I feel like this should not be in the state, but something kept in the (generic) index function. It can then be given as a parameter to `removeOld`.

src/store/albums.js
```diff
@@ -1,34 +1,49 @@
...
   },
   mutations: {
     setAlbums(state, payload) {
-      state.albums = {};
-      for (let album of payload) {
-        state.albums[album.id] = album;
+      const loaded = new Date();
+      for (let album in payload) {
+        payload[album].loaded = loaded;
```
charvp, line 16: It makes more sense to me to do this transformation in the `api` module.

src/store/commit.js
```diff
@@ -0,0 +1,33 @@
+import { indexGenerator } from "../api/fetch";
+
+export function fetchAll({ commit, rootState }, { collection, commitAction }) {
```
charvp, line 3: Passing the `index` function as a parameter would make this code more generic. Also, pass normal arguments as normal arguments. Creating an object to immediately destructure it doesn't make much sense.

src/store/commit.js
```diff
@@ -0,0 +1,33 @@
+import { indexGenerator } from "../api/fetch";
```
charvp, line 1: The name of this file doesn't feel right to me. Something like `actions` maybe?

src/store/commit.js
```diff
@@ -0,0 +1,33 @@
...
+
+export function fetchAll({ commit, rootState }, { collection, commitAction }) {
+  return new Promise((resolve, reject) => {
```
charvp, line 4: Putting the boundary of async function to promise here doesn't make much sense to me. You can just treat `fetchAll` as a Promise in the different store files if you make it an async function.

src/store/commit.js
```diff
@@ -0,0 +1,33 @@
...
+            results = {};
+          }
+          i++;
+          fetch();
+        }
+      } catch (error) {
+        reject(error);
+      }
+    }
+    fetch();
```
charvp, line 31: If `fetch` is an async function, you need to treat it as such. Either `await` it, or call `.then` on the implicit Promise it returns. (But like I said, making `fetchAll` async and `await`ing here feels more right to me.)

src/store/albums.js
robbevp: I've put this in the state, so that we can use this value (for example, to disable edit as discussed above)

src/api/fetch.js
robbevp: FYI. both work https://javascript.info/async-await#error-handling

src/store/albums.js
robbevp: (also, we need this per module, since we don't always load all modules at the same time)

src/api/fetch.js
robbevp: I've made it clearer that this condition should always be valid. on `.next()` the generator will pick up where it left of, so we need something to keep this looping until it reaches a `return` statement

robbevp:
@charvp any news on this?

src/store/albums.js
charvp: I fail to see how this would help to disable edit. For that you would need a "loading" property, not the time you started loading.

src/store/commit.js
charvp: I meant the name of `src/store/commit.js`, not `src/api/fetch.js`. Sorry for the confusion.

src/store/albums.js
robbevp: A loading property would either disable editing on all albums or none. 
My intention was to compare startLoading to loaded prop on the album. That way we can re-enable editing on an individual album from the moment it is reloaded (even while other albums are still being loaded). Thus, we disable for the shortest time possible.

charvp: Hmm, didn't think about that. Sure, no problem then.

robbevp:
> Good work, just some small comments left. Sorry for the delay in reviewing this, but as you know I've been busy.

This should be ready now

charvp:
Also, the flags count only updates after all items are fetched (per resource). Do you know why? It doesn't make a lot of sense to me, I would expect it to update every five requests.

robbevp:
> Also, the flags count only updates after all items are fetched (per resource). Do you know why? It doesn't make a lot of sense to me, I would expect it to update every five requests.
> 
> EDIT: Actually, the functionality broke because of #117, so this doesn't necessarily need to be handled here.

Can you add your findings to #117 or describe why this breaks the flag count?

robbevp:
> The loading icon stops after the first page of each resource is fetched. Is this intended? Because I would think we want it to continue until all resources are fetched?

This should be fixed now

robbevp:
🎉 🎉 🎉

charvp:
Hmm, seems like the flags issue also breaks navigation. I'll work on a fix.

PR #45, opened 2020-03-21
Improve rescan triggering

### LINKED ISSUE #22: Improve rescan triggering

robbevp:
fix #22
depends on accentor/api#17


charvp:
I would prefer two separate PRs for this.

src/views/Library.vue
```diff
@@ -12,11 +12,18 @@
...
+          >mdi-refresh
+          {{ rescan.last_click > rescan.finished_at ? "mdi-spin" : "" }}</VIcon
+        >
         {{ $t("library.start-scan") }}
       </VBtn>
     </VRow>
     <VRow class="flex-column" v-if="rescan">
+      <div>
+        <strong>{{ $t("library.last-scan") }}: </strong>
+        {{ new Date(rescan.finished_at).toLocaleString() }}
```
charvp: Use [vue-i18n's builtin datetime formatting](https://kazupon.github.io/vue-i18n/guide/datetime.html) here. The locale setting in the app might differ from the browser's.

src/views/Library.vue
```diff
@@ -12,11 +12,18 @@
     </VRow>
     <VRow v-if="rescan">
       <VBtn @click="start" color="success" class="ma-2">
-        <VIcon left>mdi-refresh {{ rescan.running ? "mdi-spin" : "" }}</VIcon>
+        <VIcon left
+          >mdi-refresh
+          {{ rescan.last_click > rescan.finished_at ? "mdi-spin" : "" }}</VIcon
+        >
```
charvp:
```suggestion
        <VIcon left>
          mdi-refresh
          {{ rescan.last_click > rescan.finished_at ? "mdi-spin" : "" }}
        </VIcon>
```
Linter is being weird.

src/views/Library.vue
```diff
@@ -12,11 +12,18 @@
     </VRow>
     <VRow v-if="rescan">
       <VBtn @click="start" color="success" class="ma-2">
```
charvp: Maybe disable the button? (Using the same condition as the spinning icon.)
You probably want to extract the condition to a computed property in that case.

src/store/rescan.js
```diff
@@ -29,6 +29,7 @@ export default {
       return start(rootState.auth)
         .then(result => {
           result.running = true;
+          result.last_click = new Date();
```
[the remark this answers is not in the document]
robbevp: This value is only set after scan is triggered, so I'm not sure why we would do this. If the user hasn't triggered a scan, this is `undefined` and will always be false in comparison to `rescan.finished_at`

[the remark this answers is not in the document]
charvp: Ah, read that wrong. No problem then. I do think setting a default value on app startup is a good idea though, because I don't like relying on comparisons between different types.

src/views/Library.vue
robbevp: have done this. The way vue-i18n handles dates isn't optimal though, it decides too much for us

charvp: In what way does it decide to much for us?

charvp: Probably because we can't do this? https://github.com/kazupon/vue-i18n/issues/625

robbevp: there seems to be little control in the order. For `en` this is `Sat, Mar 21, 2020, 4:59 PM ` even though the format is defined in the order year, month, day, ...
There is also no option to define whenther to use a comma, slash, ... in the notation

robbevp: > Probably because we can't do this? [kazupon/vue-i18n#625](https://github.com/kazupon/vue-i18n/issues/625)

indeed. Though if this is in the works, I would keep the current formatting it uses.

charvp: Yeah, I see what you mean. I do think the benefit of using our i18n library outweighs the disadvantage of having little control over the formatting.

I subscribed to (and +1'd) the issue I linked, so if it ever gets implemented we can change it to be more to our liking.

robbevp: 👍

src/store/rescan.js
```diff
@@ -3,19 +3,23 @@ import { show, start } from "../api/rescan";
...
+    rescan: null,
+    lastClick: new Date(0)
   },
   mutations: {
+    setLastClick(state, payload) {
+      state.lastClick = payload;
+    },
     setRescan(state, payload) {
-      state.rescan = payload;
+      state.rescan = Object.assign({}, state.rescan, payload);
```
charvp: I don't think this change is necessary anymore? Correct me if I'm wrong.

src/store/rescan.js
```diff
@@ -30,6 +34,7 @@ export default {
         .then(result => {
           result.running = true;
           commit("setRescan", result);
+          commit("setLastClick", new Date());
```
charvp: Can't we do this before starting the HTTP request? The change in the GUI would be more immediate I think.

src/store/rescan.js
robbevp: Thought we still needed this for reactivity, but apparently not

PR #48, opened 2020-03-21
Cache entire vuex store

### LINKED ISSUE #39: Cache DB in browser

charvp:
I still want to test the performance locally with the production api server.

Fixes #39.


charvp:
Performance is terrible while playing music, because the store is updated quite frequently then. Initial load also takes a few seconds, which is unacceptable. This will need a different approach if we want to do this.

PR #50, opened 2020-03-22
redirect back to main collection after delete on objects page

### LINKED ISSUE #40: Users should be redirected to a meaningful page when deleting a resource from its page

robbevp:
fix #40

### DIFF
--- a/src/views/albums/Album.vue
+++ b/src/views/albums/Album.vue
@@ -58,6 +58,13 @@ import AlbumArtists from "../../components/AlbumArtists";
 export default {
   name: "Album",
   components: { AlbumArtists, TracksTable, AlbumActions },
+  watch: {
+    album: function() {
+      if (this.album === undefined) {
+        this.$router.push({ name: "albums" });
+      }
+    }
+  },
   computed: {
     ...mapState("albums", ["albums"]),
     ...mapState("labels", ["labels"]),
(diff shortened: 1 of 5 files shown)


robbevp:
> Not sure if this is the behaviour we want. If a user accesses an album through an artist and then deletes the album (on the album page), I would assume we want to redirect the user back to the artist page, and not the albums page.
> 
> I think we do something similar for edits. Maybe take a look there on how its done. Though I think it uses the return to parameter which I am now realising isn't the cleanest solution. I think just popping a page of the history in the router would also work and not require extra URL params.

There are two options that I can see: 
* adding a `query.redirect` param to all possible links. (This is how it was done for forms)
* Use [`router.go(-1)`](https://router.vuejs.org/guide/essentials/navigation.html#router-go-n), which is the same as the user pressing back in the browser

charvp:
> > Not sure if this is the behaviour we want. If a user accesses an album through an artist and then deletes the album (on the album page), I would assume we want to redirect the user back to the artist page, and not the albums page.
> > I think we do something similar for edits. Maybe take a look there on how its done. Though I think it uses the return to parameter which I am now realising isn't the cleanest solution. I think just popping a page of the history in the router would also work and not require extra URL params.
> 
> There are two options that I can see:
> 
>     * adding a `query.redirect` param to all possible links. (This is how it was done for forms)
> 
>     * Use [`router.go(-1)`](https://router.vuejs.org/guide/essentials/navigation.html#router-go-n), which is the same as the user pressing back in the browser

Like I said, I think I like the latter option more.

PR #53, opened 2020-03-23
Update on this day when the date changes

### LINKED ISSUE #37: On this day doesn't change when day changes

robbevp:
fix #37


src/views/App.vue
```diff
@@ -176,6 +176,9 @@ export default {
   },
   created() {
     this.loadData();
+    setInterval(() => {
```
charvp: Save the data returned by `setInterval` and call `clearInterval` on it when the App component is destroyed. (Like here: https://github.com/dangvanthanh/vue-clock/blob/master/src/VueClock.vue#L28)

PR #54, opened 2020-03-23
Make playlist draggable

### LINKED ISSUE #13: Allow dragging in the play queue to change order

robbevp:
fix #13

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "vue-headful": "^2.1.0",
     "vue-i18n": "^8.15.6",
     "vue-router": "^3.1.6",
+    "vuedraggable": "^2.23.2",
     "vuetify": "^2.2.18",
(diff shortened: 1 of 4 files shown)


robbevp:
> Looks good, but I think we want a bit more of a visual hint that the rows are draggable. Something like `drag-horizontal-variant` icon somewhere in the row might be a good addition. (I would move the remove icon to the end of the row and place the drag icon at the start.)
> 
> We might also want to think about only making that icon draggable, if that's doable with the library you added. (But when dragging the icon, the whole row should be dragged.)

Agreed with the icon.
I'm not sure about limiting the dragging action to the icon. The current implementation doesn't block any of the other links or actions in the row. I thought about this, but when testing it just didn't feel necessary. (But maybe after using this for a while I might feel different about it)

PR #58, opened 2020-03-24
Base order on the current length of track_artists

### LINKED ISSUE #34: Newly created artists aren't added in the right order on mass edit

robbevp:
fix #34


src/components/MassEditDialog.vue
```diff
@@ -419,7 +419,7 @@ export default {
                     artist_id: id,
                     name: ta.name || ta.artist_id,
                     role: ta.role,
-                    order: index + 1,
+                    order: index,
                   });
```
charvp: As discussed, move this (implicit -1) to line 504.

PR #67, opened 2020-03-27
Make sure getMonth and getDate always have zero padding to compare to iso string

robbevp:
Didn't spot this before but `.getMonth` doesn't provide zero padding

### DIFF
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -122,7 +122,7 @@ export default {
         .filter(
           (r) =>
             `${r.release.slice(-5)}` ===
-            `${("0" + (today.getMonth() + 1)).slice(-2)}-${today.getDate()}`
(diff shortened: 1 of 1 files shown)


src/store/albums.js
```diff
@@ -121,8 +121,9 @@ export default {
       return getters.albums
         .filter(
           (r) =>
-            `${r.release.slice(-5)}` ===
-            `0${today.getMonth() + 1}-${today.getDate()}`.slice(-5)
+            `${r.release.slice(-5, -3)}` ===
+              `0${today.getMonth() + 1}`.slice(-2) &&
```
charvp: Use [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart).

PR #71, opened 2020-03-28
Make album and artists cards have a fixed height

### LINKED ISSUE #19: Make album and artist cards have all the same height

robbevp:
fix #19


src/components/AlbumCard.vue
```diff
@@ -2,13 +2,19 @@
   <VCard :to="{ name: 'album', params: { id: album.id } }">
     <VImg :aspect-ratio="1" :src="album.image500" v-if="album.image500" />
     <VImg :aspect-ratio="1" :src="album.image" v-else-if="album.image" />
-    <VCardTitle class="pb-0">
+    <VImg
+      :aspect-ratio="1"
+      :src="require('../assets/svg/album.svg')"
```
charvp: I would prefer not to bring parts of the MDI dependency in-tree. Is there no way to use an icon here or get this SVG from the yarn dependency?

src/sass/_clamp.scss
```diff
@@ -0,0 +1,19 @@
+.clamp {
+  display: -webkit-box;
```
charvp: Doesn't this mean only webkit browser would support this functionality?

robbevp: https://caniuse.com/#feat=css-line-clamp

src/components/AlbumCard.vue
robbevp: The dependency didn't include the SVGs. Since the icons behave as text, using that made it a lot more complex for a lesser result

charvp: https://www.npmjs.com/package/@mdi/svg?

src/sass/_clamp.scss
charvp: > As there is no specification and the property is dependent on an outdated implementation of flexbox (hence display: -webkit-box), an alternative solution may at some point replace it.

charvp: Not a big fan of that.

robbevp:
> I think this could be a lot less complex if we just make it one line of content and use browser tooltips to make sure users can still see the full content.

Only one line of content is fine for me. This also means that we can use the `white-space` property instead of `line-clamp`. Using a tooltip instead of simply expanding seems more complex to me

PR #75, opened 2020-04-04
Expose merging of genres

robbevp:
Not sure if we want to warn the user of the consequences of their actions, since this is restricted to mod+


src/api/genres.js
```diff
@@ -98,3 +98,20 @@ export function destroyEmpty(auth) {
         : request.json().then((result) => Promise.reject(result));
     });
 }
+
+export function merge(auth, oldID, newID) {
```
charvp, line 102: I think switching the order of these arguments makes sense. For people who know how the API is implemented, that order is the most logical order (or at least it is to me).

src/api/tracks.js
```diff
@@ -49,6 +49,22 @@ export function create(auth, track) {
     });
 }
 
+export function read(auth, id) {
```
charvp, line 52:
```suggestion
export function get(auth, id) {
```

src/store/tracks.js
```diff
@@ -47,6 +47,24 @@ export default {
...
+          this.commit("addError", error);
+          return Promise.resolve(false);
+        });
+    },
+    readMultiple({ dispatch }, ids) {
+      const promises = [];
+      ids.forEach((id) => {
+        promises.push(dispatch("read", id));
+      });
+      return Promise.all(promises);
```
charvp, line 66: I think we will need to handle these requests sequentially. This could potentially launch thousands of requests at the same time.

robbevp: We don't want to launch that many requests period.
One of the reasons why I created accentor/api#39

robbevp: We could also get all the tracks in the scope of the genre that is merged into. This means fetching more data than we stricly need, but it means that we can keep the number of requests low.
(That would still not solve the problem after a destroy, though)

robbevp:
@charvp I've updated this as discussed

src/api/tracks.js
```diff
@@ -49,6 +49,22 @@ export function create(auth, track) {
     });
 }
 
+export function get(auth, id) {
```
charvp, line 52: Is this still used?

src/store/tracks.js
```diff
@@ -47,6 +66,17 @@ export default {
           return Promise.resolve(false);
         });
     },
+    read({ commit, rootState }, id) {
```
charvp, line 69: Is this still used?

src/store/tracks.js
```diff
@@ -23,6 +23,25 @@ export default {
...
       Vue.delete(state.tracks, id);
     },
+    updateGenreOccurence(state, { newID, oldID }) {
+      for (const t in state.tracks) {
+        const i = state.tracks[t].genre_ids.findIndex(
+          (gId) => `${gId}` === `${oldID}`
+        );
+        if (i >= 0) {
+          state.tracks[t].genre_ids.splice(i, 1);
+          // if newID isn't provided we'll just remove the old genre
```
charvp, line 33: I feel like we shouldn't handle this case? If newID isn't provided, it's a programmer error to use the function in the wrong way.

src/components/GenreActions.vue
```diff
@@ -15,8 +15,61 @@
     >
       <VIcon>mdi-pencil</VIcon>
     </VBtn>
+    <VDialog
```
charvp, line 18: Can we extract this to its own component?

src/store/tracks.js
robbevp: This way, we can use this function after a genre merge and after a genre is deleted (see #77 )
I haven't implemented that yet, since that seemed like something for a different PR

charvp: If that is something we want to do, I feel like it should be a different function or this one should be renamed (when it is implemented). Until then, it doesn't make sense to me to already handle this.

robbevp: Sure

src/components/GenreMergeDialog.vue
```diff
@@ -0,0 +1,94 @@
...
+      </VCardTitle>
+      <VCardText>
+        <VContainer>
+          <VRow>
+            <VCol cols="12">
+              <VCombobox
+                :items="sortedGenres"
+                cache-items
+                chips
+                deletable-chips
```
charvp, line 35: Tested this locally. I feel like these shouldn't be chips? If I understand correctly, we only want to be able to select one genre here, so a regular select seems more appropriate.

robbevp: you're right - i've updated

PR #81, opened 2020-04-10
Make sure player doesn't get stuck on empty tracks

### LINKED ISSUE #80: Skip track without audio in playlist
When adding tracks to the playlist, track without audio are not added. There are still two ways the player can get stuck on empty tracks:
* [ ] A track might be updated in the time between adding it to the playlist and it being the current track 
* [ ] The audiofile might have been removed, but the server and/or the store is not yet aware of this

robbevp:
fix #80

### DIFF
--- a/src/components/AlbumActions.vue
+++ b/src/components/AlbumActions.vue
@@ -64,7 +64,7 @@ export default {
     ...mapGetters("auth", ["isModerator"]),
     tracks() {
       const getter = this.$store.getters["tracks/tracksFilterByAlbum"];
-      return getter(this.album.id).map((t) => t.id);
+      return getter(this.album.id);
     },
   },
   methods: {
@@ -75,7 +75,7 @@ export default {
       }
     },
     startTracks: function () {
-      const queue = this.tracks.filter((track) => track.length !== null);
+      const queue = this.filterTracks();
       if (queue.length > 0) {
         this.$store.commit("player/playTracks", queue);
       } else {
@@ -85,7 +85,7 @@ export default {
       }
     },
     addTracks: function () {
-      const queue = this.tracks.filter((track) => track.length !== null);
+      const queue = this.filterTracks();
       if (queue.length > 0) {
         this.$store.commit("player/addTracks", queue);
       } else {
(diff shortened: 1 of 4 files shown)


src/components/AlbumActions.vue
```diff
@@ -100,6 +100,14 @@ export default {
         newAlbum: { review_comment: reviewComment },
       });
     },
+    filterTracks: function () {
```
charvp, line 103: Write this as a filter + map, that's much clearer. I also don't think this needs to be a separate function, the filter plus map operation is easy enough. (Or at least the function needs to be named better.)

src/components/Player.vue
```diff
@@ -208,9 +208,11 @@ export default {
     doSeek() {
       if (this.doSeek) {
         this.$store.commit("player/setDoSeek");
-        this.$refs.audio.currentTime = this.seekTime;
-        if (this.playing) {
-          this.$refs.audio.play();
+        if (Number.isFinite(this.seekTime)) {
```
charvp, line 211: Why?

src/components/Player.vue
```diff
@@ -306,6 +308,11 @@ export default {
         this.setSeekTime(time);
       }
     },
+    onAudioError(event) {
+      if (event.srcElement.networkState === 3) {
```
charvp, line 312: The server will return a 404 if there is no audio file attached. I would only skip in that case, since if it is another error, the error is much more catastrophic (and probably every track will be skipped). Also let the user know that an error happened (and disambiguate between the case where there is just no audio file and we skipped to the next track and the case where a much more catastrophic error happened).

src/components/Player.vue
robbevp: This callback gets checked before `onAudioError` is called and kept producing a ton of error messages in the console. This isn't necessary per se.

src/components/Player.vue
robbevp: The problem is that the `onerror` event doesn't include the http response code. The only to way to know the server returned a 404 is by launching the event again, which doesn't seem like a clean solution to me

robbevp: The more catastrophic error would either be (1) the client is not connected to the internet (2) the server is down. Maybe we should alert the user if either of those is the case and prevent playing (since it wouldn't make sense)?

Issue #154, opened 2020-06-09
Track isn't updated after edit


robbevp:

robbevp:
We receive the correct json object from the api, but nothing is done with them

robbevp:
@charvp I think the caching in `trackBinnedBy...` has something to do with this.
An updated tracks shows up correctly if you reopen the edit form and in the playlist, but not in the `trackTable`

PR #155, opened 2020-06-12
Fix performance issues

### LINKED ISSUE #148: Figure out and fix memory leak
Not necessarily introduced by #44, but it certainly exacerbated the issue. Before base memory usage (on the settings page) was ~80MiB, now it is around ~800MiB.

robbevp:
fix #148 #154

### DIFF
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -44,21 +44,19 @@ export default {
         }
       }
     },
-    updateLabelOccurence(state, { newID, oldID }) {
-      for (const a in state.albums) {
-        const i = state.albums[a].label_ids.findIndex(
-          (lId) => `${lId}` === `${oldID}`
-        );
+    updateLabelOccurence(state, { oldID, newID }) {
+      const oldAlbums = state.albums;
+      state.albums = {};
+      for (let album of Object.values(oldAlbums)) {
+        const i = albums.album_labels.findIndex((l) => l.label_id === oldID);
         if (i >= 0) {
-          state.albums[a].label_ids.splice(i, 1);
-          if (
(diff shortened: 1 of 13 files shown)


robbevp:
Todo:
* [ ] convert `setObject` mutation to new system
* [ ] convert `updateOccurences` to replace whole object

robbevp:
@charvp I've tested this with some edits for various types of objects. There is a small spike in memory usage after a mass edit (tested with 40 tracks), but the garbage collector is able to handle this after a while.

charvp:
@robbevp Can you take a look at my commit. I'll approve and merge once I've heard something from you.

src/store/albums.js
```diff
@@ -44,21 +44,19 @@ export default {
         }
       }
     },
-    updateLabelOccurence(state, { newID, oldID }) {
-      for (const a in state.albums) {
-        const i = state.albums[a].label_ids.findIndex(
-          (lId) => `${lId}` === `${oldID}`
-        );
+    updateLabelOccurence(state, { oldID, newID }) {
+      const oldAlbums = state.albums;
```
robbevp, line 48: Is there a reason why this doesn't use the spread operator? I would want to keep all the commit as similar as possible

robbevp: If it isn't necessary to use a spread operator, we should remove it for all `const oldObjects = `

charvp: Because it's not necessary? No need to create a new (large) object if it's not required.

charvp: > If it isn't necessary to use a spread operator, we should remove it for all `const oldObjects = `

Sure, agreed.

PR #156, opened 2020-06-12
Update related objects after destroy

### LINKED ISSUE #77: Update related objects after destroy
If the user deletes an object, the related objects are not updated. This leads to small errors and unexpected behaviour (eg. components that just don't show up).

## A first list of the objects to update
On album destroy:
_Nothing?_

On artist destroy:
* Tracks where artist is in track_artists
* Albums where artist is in album_artists

On genre destroy:
* Tracks where genre is in track_genres

On label destroy:
* Albums where label is in album_labels

On track destroy:
_Nothing?_

robbevp:
fix #77

### DIFF
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -44,13 +44,16 @@ export default {
         }
       }
     },
-    updateLabelOccurence(state, { oldID, newID }) {
+    removeOrMergeLabelOccurence(state, { oldID, newID }) {
       const oldAlbums = state.albums;
       state.albums = {};
       for (let album of Object.values(oldAlbums)) {
         const i = albums.album_labels.findIndex((l) => l.label_id === oldID);
         if (i >= 0) {
-          if (album.album_labels.some((l) => l.label_id === newID)) {
+          if (
+            typeof newID !== undefined &&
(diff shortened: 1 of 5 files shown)


robbevp:
Todo:
* [ ] convert `setObject` mutation to new system
* [ ] convert `updateOccurences` to replace whole object

robbevp:
@charvp reworked after #155 - should be ready now

src/store/albums.js
```diff
@@ -44,13 +44,27 @@ export default {
...
+      state.albums = {};
+      for (let album in Object.values(oldAlbums)) {
+        const i = album.album_artists.findIndex((aa) => aa.artist_id === oldID);
+        if (i >= 0) {
+          album.album_artists.splice(i, 1);
+        }
+        state.albums[album.id] = album;
+      }
+    },
+    removeOrMergeLabelOccurence(state, { oldID, newID }) {
```
charvp, line 58: I'm not sure about doing the remove and merge actions in the same function. There's some shared boilerplate, but I would still split these up. (Same for genre.)

robbevp: It felt like too much boilerplate to me, but sure.

PR #157, opened 2020-06-13
Trigger timeout if rescan is running

### LINKED ISSUE #138: Add interval for rescan if rescan is currently running
If a user loads or reloads the data and a rescan is running, the state of the rescan should be set to running and the info should be fetched on the same interval as for the user that triggered it.

robbevp:
fix #138

### DIFF
--- a/src/store/rescan.js
+++ b/src/store/rescan.js
@@ -19,7 +19,10 @@ export default {
       return show(rootState.auth)
         .then((result) => {
           commit("setRescan", result);
-          if (rootState.rescan.lastClick > new Date(result.finished_at)) {
+          if (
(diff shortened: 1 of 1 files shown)


src/store/rescan.js
```diff
@@ -19,7 +19,10 @@ export default {
       return show(rootState.auth)
         .then((result) => {
           commit("setRescan", result);
-          if (rootState.rescan.lastClick > new Date(result.finished_at)) {
+          if (
+            rootState.rescan.lastClick > new Date(result.finished_at) ||
+            Boolean(result.running)
```
charvp, line 24: Why the boolean cast? Seems unnecessary?

PR #169, opened 2020-06-15
Expose tracks merge functionality

### LINKED ISSUE #15: Track merge functionality should be exposed somehow

robbevp:
fix #15

### DIFF
--- a/src/locales/nl.json
+++ b/src/locales/nl.json
@@ -284,13 +284,17 @@
 			"edit": "Pas nummer aan",
 			"empty": "Dit nummer heeft geen audio.",
 			"length": "Lengte",
+			"merge": "Voeg nummer samen",
+			"merge-explanation": "Dit verwijderd dit nummer en de bijhorende tags maar bewaard het audiobestand indien aanwezig.",
+			"merge-into": "Voeg {obj} samen met",
 			"number": "Tracknummer",
 			"update": "Werk nummer bij"
 		},
 		"tracks": "Nummers | Nummer | Nummers"
 	},
 	"page-titles": {
-		"edit": "{obj} aanpassen"
+		"edit": "{obj} aanpassen",
+		"merge": "{obj} samenvoegen"
 	},
 	"search": "Zoeken",
 	"users": {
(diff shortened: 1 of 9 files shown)


robbevp:
@charvp Might be good idea to discuss the layout of MergeTrack over the weekend, since this is more complex than the merging of genres or labels

robbevp:
@charvp Some cleanup and changes will be necessary, but this is ready for a first review.

Some notes:
* Having the table for comparison appear and dissapear when a track was selected and deselected was a bit too much, so I made sure we can always display this table.
* Swap tracks only changes the result. I was thinking about switching the columns of the table, but wasn't sure if it was necessary.

robbevp:
@charvp I've been thinking about your idea of showing a diff-like view.
Another way to show the tracks and result would be to have the result in the middle and show arrows to make it clear from which track each property comes.

robbevp:
>     * I would try to show in the first two columns which property is used as well. Make it green or something like that.
I didn't add color, but went with arrows. Using colors would make the text less readable (and we would need another signifier anyway for accessibility).  If you feel this isn't clear enough, I would put the result column in the center

Other stuff is fixed.

src/locales/nl.json
```diff
@@ -282,13 +282,26 @@
 		"track": {
 			"empty": "Dit nummer heeft geen audio.",
 			"length": "Lengte",
+			"merge": {
+				"explanation": "Dit verwijderd dit nummer en de bijhorende tags maar bewaard het audiobestand indien aanwezig.",
```
charvp, line 286:
```suggestion
				"explanation": "Dit verwijdert dit nummer en de bijhorende tags maar bewaart het audiobestand indien aanwezig.",
```

robbevp:
> The arrows are a little unclear to me. I would put the result in the middle and point right and left from the outer columns. Also, the arrows seems wrong? If you look at the following screenshot, the metadata of the selected track is used and the audiofile of the original track, but all the arrows are in the first column:
> screenshot_2020-07-25-114844
> 
> Also this should probably also depend on #215.

Agreed, I was thinking about switched those columns a while ago.
All other stuff is fixed. Once #215 is merged, I'll rework this to fit in with the new `TrackActions`

PR #181, opened 2020-06-20
Add form validations

### LINKED ISSUE #100: Validate forms clientside
A lot of validations can happen clientside.
We should try and do as many as possible after input instead of after submit

robbevp:
fix #100 

depends on #182

Uses [vuetify's built-in validation](https://vuetifyjs.com/en/components/forms/)
* If a field only required one rule, I've added it directly as an attribute
* If a field required multiple rules or more complex, I've added it to a computed property (one per form). 
* Lazy validation is used when validating too eagerly would be annoying (fe. password confirmation) or if the form is not on a specific page (Forms on library pages).

### DIFF
--- a/src/components/CodecConversionForm.vue
+++ b/src/components/CodecConversionForm.vue
@@ -1,16 +1,20 @@
 <template>
-  <VForm>
+  <VForm v-model="isValid" ref="form" lazy-validation>
     <VRow>
       <VCol cols="3">
         <VTextField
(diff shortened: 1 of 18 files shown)


robbevp:
@charvp was there a specific reason why the `MassEditDialog` doesn't use a `VForm` component? It would be handy if we could do the validations the same way as in all the other forms (and that needs `VForm`)

PR #204, opened 2020-07-04
Expose track details to mod+

### LINKED ISSUE #33: Allow moderators to see audio file details

robbevp:
fix #33


robbevp:
@charvp I've updated this based on our discussion. 
This already includes support for accentor/api #74, but should work before that is merged or while we are still updating all the audio files

robbevp:
> The implementation looks fine, but I'm not a fan of how it looks. The tracks table is getting very busy. I propose moving edit and delete to a submenu and moving the track info to there as well. The track info will probably have to be changed to a full popup to still make things look OK, but I don't see any problem with that. The actual track info view itself could also be a bit prettier; just a key value list isn't very nice to look at.

I've reworked the trackactions in a separate PR #215

charvp (changes requested):
Looks good! Only thing I would change is moving it to the top of the menu and giving the icon a colour (info blue seems the most appropriate to me).

robbevp:
> Looks good! Only thing I would change is moving it to the top of the menu and giving the icon a colour (info blue seems the most appropriate to me).

I'm not sure about the order, but don't feel very strongly. I assume that using this will quickly show a logical order

charvp (approved):

PR #208, opened 2020-07-08
Refactor EditAlbum and NewAlbum

robbevp:
Refactors the EdiAlbum and NewAlbum views to use one component with a form for both. This will prevent both views getting out of sync in functionality.

fix #207

PR #209, opened 2020-07-09
Refactor EditArtist and NewArtist to use shared form

robbevp:
Same as #208


robbevp:
> Did you test this? Both creating a new artist and editing an existing one?

Yes - tested for both

PR #210, opened 2020-07-09
Refactor user forms

robbevp:
Refactor Settings, EditUser, NewUser to use a shared form

### DIFF
--- a/src/views/settings/Settings.vue
+++ b/src/views/settings/Settings.vue
@@ -20,7 +20,7 @@
     </VRow>
     <VRow>
       <VCol>
-        <AuthTokensTable :authTokens="authTokens" redirectFallback="home" />
+        <AuthTokensTable :authTokens="authTokens" />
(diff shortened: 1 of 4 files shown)


src/views/settings/Settings.vue
```diff
@@ -15,69 +15,27 @@
...
-            {{ $t("common.change-settings") }}
-          </VBtn>
-        </VForm>
+        <UserForm :user="user" redirectFallback="home" />
       </VCol>
     </VRow>
     <VRow>
       <VCol>
-        <AuthTokensTable :authTokens="authTokens" />
+        <AuthTokensTable :authTokens="authTokens" redirectFallback="home" />
```
charvp, line 23: `AuthTokensTable` doesn't take this property (and it shouldn't, since it never redirects).

robbevp:
> Same question here: did you test all impacted functionality?

Jep

PR #227, opened 2020-07-26
Prevent watch from overwriting local changes

### LINKED ISSUE #217: Forms are reset after each update to the collection
Because we "reset" the whole collection (tracks/albums/artists/...) after most updates to the store, we update each track/album/artist/... quite often. (Especially while loading all data).

Since (almost) all forms contain a `watch` property for their main object, this means that the content of the form is reset after each update to the collection (even if it has nothing to do with the specific object you are editing). Meaning you lose whatever edits you have made and not yet saved.

I propose to add an `isDirty` property to each form. If the form is dirty we should not run `fillValues`. (This would also help with #31 )

robbevp:
fix #217 

This will also help with #31

### DIFF
--- a/src/components/AlbumForm.vue
+++ b/src/components/AlbumForm.vue
@@ -1,6 +1,6 @@
 <template>
   <VRow no-gutters align="center" justify="center">
-    <VCol lg="6" sm="8" cols="12">
+    <VCol lg="6" sm="8" cols="12" @change.once="isDirty = true">
       <VAlert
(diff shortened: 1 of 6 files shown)


charvp (approved):
Just wondering: why does the change handler have to be installed on the parent of the form?

robbevp:
This had been a while, so I don't exactly remember. I think it was that vuetify's VForm didn't have a change-event which could be used

PR #237, opened 2020-08-02
Allow mod+ to review empty tracks

### LINKED ISSUE #226: Allow mod+ to view all tracks without an audio file attached
In the library setting we have a button to handle empty artists, albums, genres and labels. Since we can now merge tracks we should add a button to handle empty tracks. This should lead the mod+ to a filtered TrackTable with all tracks without an audio file attached.

robbevp:
fix #226 

I was doubting to make this page a subpage of tracks or library. I went for tracks, but don't feel very strongly about it.

### DIFF
--- a/src/components/MaintenanceActions.vue
+++ b/src/components/MaintenanceActions.vue
@@ -1,13 +1,21 @@
 <template>
   <span>
-    <VBtn
-      :to="{ name: 'tracks-without-audio' }"
-      color="info"
-      class="ma-2 white--text"
-    >
-      <VIcon left>mdi-alert-octagon</VIcon>
-      {{ $t("library.overview-tracks-without-audio") }}
-    </VBtn>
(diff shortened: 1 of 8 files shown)


charvp (changes requested):
As discussed in person, I'm not a huge fan of the terminology used here. Instead of the verb review, I would just say overview of and instead of empty tracks I would use tracks without audio. Developers might know what empty tracks are, but I don't think this is terminology that would be clear on first use.

Also, the colour of the button shouldn't be red, since clicking it isn't a destructive/dangerous action. I would opt for info blue.

Can we add a count to the button? And maybe hide it if that count is 0?

robbevp:
Agreed on naming and color.

I wouldn't add a count to the button, since i would like to keep this button in style with the others. (So adding a count would need to happen in a separate PR for all maintance actions.)
I have disabled the button if there are no tracks without audio in the library

charvp (approved):

PR #268, opened 2020-09-13
Switch order of newTrack and oldTrack in TrackMerge

robbevp:
This PR switched the default order of the newTrack and oldTrack in the merge view.

Since we _usually_ start from a track without audio (especially with the new overview), it makes more sense to have the keep the tags from the track that is the starting point and keep the audio from the track that is selected within the merge view. (Previously, the default would be the other way around, if you select a track without audio, this would make that you wouldn't not keep anything from the oldTrack)

### DIFF
--- a/src/views/tracks/MergeTrack.vue
+++ b/src/views/tracks/MergeTrack.vue
@@ -27,11 +27,11 @@
           </thead>
           <tbody>
             <tr>
-              <td :class="{ 'd-flex justify-space-between': reversed }">
+              <td :class="{ 'd-flex justify-space-between': !reversed }">
                 <span class="my-auto"
                   >{{ track.number }}. {{ track.title }}</span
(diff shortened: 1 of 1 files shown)


src/views/tracks/MergeTrack.vue
```diff
@@ -268,8 +268,8 @@ export default {
       this.newID = id;
     },
     submit() {
-      const newID = this.reversed ? this.track.id : this.newID;
-      const oldID = this.reversed ? this.newID : this.track.id;
+      const newID = this.reversed ? this.newID : this.track.id;
+      const oldID = this.reversed ? this.track.id : this.track.id;
```
charvp, line 272:
```suggestion
      const oldID = this.reversed ? this.track.id : this.newID;
```
?

PR #275, opened 2020-09-22
Use generic functions for all API calls

robbevp:
Refactor to use the generic create and read functions from `fetch.js` for the rescan start and show api calls and creating auth_tokens.

### DIFF
--- a/src/api/auth_tokens.js
+++ b/src/api/auth_tokens.js
@@ -1,23 +1,15 @@
-import baseURL from "./base_url";
-import { indexGenerator, destroy as genDestroy } from "./fetch";
+import {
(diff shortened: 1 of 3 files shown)


robbevp:
Switched back to draft, since I would do #279 first, and then rework this

robbevp:
Since a `Request` [doesn't allow custom properties](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) and I didn't want every API call to user `fetch-retry`, I moved away from providing `resolveRequest` with a `Request` and provided it the path and the options.

charvp (approved):

PR #276, opened 2020-09-22
Fetch object before allowing edits

### LINKED ISSUE #73: Fetch object when opening editing form
For most of the main collections we should fetch the object before we allow a user to edit.
This way we prevent a user from overwriting data that they didn't see

robbevp:
fix #73 
To prevent the user from overwriting changes they didn't see, we fetch the object when opening the editing form.

### DIFF
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -123,13 +123,16 @@ export default {
           return Promise.resolve(false);
         });
     },
-    async read({ commit, rootState }, id) {
-      try {
-        const album = await read(rootState.auth, id);
-        commit("setAlbum", { id, album });
-      } catch (error) {
-        this.commit("addError", error);
-      }
+    read({ commit, rootState }, id) {
+      return read(rootState.auth, id)
+        .then((result) => {
+          commit("setAlbum", { id, album: result });
+          return Promise.resolve(result.id);
+        })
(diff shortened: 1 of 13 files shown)


src/store/albums.js
```diff
@@ -116,6 +123,14 @@ export default {
           return Promise.resolve(false);
         });
     },
+    async read({ commit, rootState }, id) {
```
charvp, line 126: I don't think we want to mix async syntax and promise syntax in the same file.

charvp (changes requested):

src/store/albums.js
robbevp: We already do this in a few files, for example: `src/api/fetch.js` and `src/store/tracks.js`
In time I would want to switch completely to async/await, but would not hold back using more modern syntax now in new additions/changes. It wil be easier to switch with gradual changes.

charvp: I agree with switching to async await completely in time. I was not aware of any files mixing the two methods of doing things, but that doesn't mean we should start doing that more, since that's generally considered bad style.

robbevp: sure, I'll edit and create an issue for this switch

charvp (approved):

Issue #114, opened 2020-05-07
Clicking play on playlist that ended should restart playlist


robbevp:
Maybe also disable button if no playlist is present?

PR #285, opened 2020-10-03
Create play after track has finished

robbevp:
depends on accentor/api#114

There is a good argument to create a play when we reach 75%, 80%, 90%, ... of the tracks length, but for the moment I've just attached to to the `ended`-event.

I put the actions inside plays, since play data will live inside the track and not get a separate module.

### DIFF
--- a/src/api/plays.js
+++ b/src/api/plays.js
@@ -1,9 +1,9 @@
 import { indexGenerator, create as genCreate } from "./fetch";
 
 export function index(auth) {
-  return indexGenerator("albums", auth);
+  return indexGenerator("plays", auth);
 }
 
 export function create(auth, play) {
-  return genCreate(`plays`, auth, { play });
+  return genCreate("plays", auth, { play });
 }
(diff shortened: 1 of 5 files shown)


src/api/plays.js
```diff
@@ -0,0 +1,9 @@
+import { indexGenerator, create as genCreate } from "./fetch";
+
+export function index(auth) {
+  return indexGenerator("albums", auth);
```
chvp, line 4: `albums`?

src/api/plays.js
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export function create(auth, play) {
+  return genCreate(`plays`, auth, { play });
```
chvp, line 8:
```suggestion
  return genCreate("plays", auth, { play });
```

chvp (changes requested):

robbevp:
Rebased and fixed typo's

chvp (approved):

Issue #279, opened 2020-09-26
Switch to async/await syntax instead of promises


robbevp:
re #276

charvp:
This can be closed?

PR #286, opened 2020-10-04
Switch promise to async syntax for store and API

robbevp:
re #279 
(I would not yet close this issue, as we still use Promises inside some of the vue templates - we might also want to switch there?)

For the store I've made two additional adjustments, to get more consistent actions:
1. An action always return something, either an id (in case of create) or true or false. (Before we had some that returned true or false and some that didn't)
2. Avoid using the global `this` of the store: `dispatch` and `commit` are always used from the actions context, with `{ root: true }` where necessary. This IMHO more clear (and we were using both at the same time

### DIFF
--- a/src/api/fetch.js
+++ b/src/api/fetch.js
@@ -4,9 +4,9 @@ const fetchRetry = require("fetch-retry")(fetch);
 export async function* indexGenerator(path, auth) {
   let page = 1;
   while (true) {
-    let request;
+    let response;
     try {
-      request = await fetchRetry(`${baseURL}/${path}?page=${page}`, {
+      response = await fetchRetry(`${baseURL}/${path}?page=${page}`, {
         retries: 5,
         retryDelay: function (attempt) {
           return Math.pow(2, attempt) * 15000;
@@ -20,13 +20,13 @@ export async function* indexGenerator(path, auth) {
     } catch (reason) {
       throw { error: [reason] };
     }
-    const result = await request.json();
-    if (request.ok && result) {
+    const result = await response.json();
+    if (response.ok && result) {
       const loaded = new Date();
       for (let obj in result) {
         result[obj].loaded = loaded;
       }
-      if (request.headers.get("x-total-pages") == page) {
+      if (response.headers.get("x-total-pages") == page) {
         return result;
       } else {
         yield result;
--- a/src/api/fetch.js
+++ b/src/api/fetch.js
@@ -4,19 +4,22 @@ const fetchRetry = require("fetch-retry")(fetch);
 export async function* indexGenerator(path, auth) {
   let page = 1;
   while (true) {
-    const request = await fetchRetry(`${baseURL}/${path}?page=${page}`, {
-      retries: 5,
-      retryDelay: function (attempt) {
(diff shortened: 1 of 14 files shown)


src/api/fetch.js
```diff
@@ -35,101 +35,86 @@ export async function* indexGenerator(path, auth) {
...
+  });
+  const result = await response.json();
+  if (response.ok) {
+    return result;
+  } else {
+    throw result;
+  }
+}
+
 export function create(path, auth, object) {
```
charvp, line 50: All of these functions should be marked as async and `return await resolveRequest(...)`, IMO. Now, looking at just the function, you wouldn't know that it returns asynchronously.

src/api/fetch.js
```diff
@@ -35,101 +35,86 @@ export async function* indexGenerator(path, auth) {
   }
 }
 
+async function resolveRequest(request) {
+  const response = await fetch(request).catch((reason) => {
```
charvp, line 39: `.catch()` is Promise-syntax.

charvp (changes requested):

src/api/fetch.js
```diff
@@ -35,101 +38,89 @@ export async function* indexGenerator(path, auth) {
...
+  } catch (reason) {
+    throw { error: [reason] };
+  }
+  const result = await response.json();
+  if (response.ok) {
+    return result;
+  } else {
+    throw result;
+  }
+}
```
charvp, line 54:
```suggestion
  try {
    const response = await fetch(request);
    const result = await response.json();
    if (response.ok) {
      return result;
    } else {
      throw result;
    }
  } catch (reason) {
    throw { error: [reason] };
  }
}
```
This is a bit cleaner, IMO.

src/api/fetch.js
```diff
@@ -4,26 +4,29 @@ const fetchRetry = require("fetch-retry")(fetch);
...
-      retryDelay: function (attempt) {
-        return Math.pow(2, attempt) * 15000;
-      },
-      method: "GET",
-      headers: {
-        "x-secret": auth.secret,
-        "x-device-id": auth.device_id,
-      },
-    }).catch((reason) => {
+    let response;
```
charvp, line 7: The same should be done here, but GitHub won't let me edit a section this large.

charvp (changes requested):

src/api/fetch.js
robbevp: This doesn't work.
If we have an error from the server (say an album without title) we will throw iit on line 48 (your version), then catch it and wrap it in an error object on line 51. As a result our errors don't display correctly anymore (but as `errors.[object object]`)

src/api/fetch.js
robbevp: same as above, this doesn't work

src/api/fetch.js
charvp: You're right. Ugh, that's ugly, but fine.

charvp:
Can you rebase to make sure that all linting errors are fixed?

robbevp:
> Can you rebase to make sure that all linting errors are fixed?

done

charvp (approved):

PR #287, opened 2020-10-04
Switch to async syntax in vue SFC

robbevp:
re #279 
See #286 for the same syntax switch in the store and API calls.

This PR switches to async/await syntax inside methods (computed and watched properties do not support this syntax).

There are some places where I've used `throw false`, as a replacement for `Promise.reject()`. We could throw any value to get the same result (but have to use *some* value), `false` made the most sense to me.

### DIFF
--- a/src/components/AlbumForm.vue
+++ b/src/components/AlbumForm.vue
@@ -359,7 +359,7 @@ export default {
         this.newAlbum.album_artists.splice(index, 1)[0]
       );
     },
-    submit() {
+    async submit() {
       const transformed = {
         title: this.newAlbum.title,
         release: this.newAlbum.release,
@@ -375,52 +375,44 @@ export default {
         album_artists: [],
       };
 
-      const promises = [];
-
-      for (let label of this.newAlbum.album_labels) {
+      const label_map = this.newAlbum.album_labels.map(async (label) => {
         if (typeof label.label_id === "string") {
-          promises.push(
-            this.createLabel({ name: label.label_id }).then((id) => {
-              if (id) {
-                transformed.album_labels.push({
-                  label_id: id,
-                  catalogue_number: label.catalogue_number,
-                });
-              } else {
-                return Promise.reject();
-              }
-            })
-          );
+          const id = await this.createLabel({ name: label.label_id });
+          if (id) {
+            transformed.album_labels.push({
+              label_id: id,
+              catalogue_number: label.catalogue_number,
+            });
+          } else {
+            throw false;
+          }
         } else {
           transformed.album_labels.push({
             label_id: label.label_id.id,
             catalogue_number: label.catalogue_number,
           });
         }
-      }
+      });
 
-      this.newAlbum.album_artists.forEach((aa, index) => {
+      const artist_map = this.newAlbum.album_artists.map(async (aa, index) => {
         if (typeof aa.artist_id === "string") {
-          promises.push(
-            this.createArtist({
-              name: aa.artist_id,
-              review_comment: "New artist",
-            }).then((id) => {
-              if (id) {
-                transformed.album_artists.push({
-                  artist_id: id,
-                  name: aa.name || aa.artist_id,
-                  separator:
-                    index !== this.newAlbum.album_artists.length - 1
-                      ? aa.separator
-                      : null,
-                  order: index + 1,
-                });
-              } else {
-                return Promise.reject();
-              }
-            })
-          );
+          const id = await this.createArtist({
+            name: aa.artist_id,
+            review_comment: "New artist",
+          });
+          if (id) {
+            transformed.album_artists.push({
+              artist_id: id,
+              name: aa.name || aa.artist_id,
+              separator:
+                index !== this.newAlbum.album_artists.length - 1
+                  ? aa.separator
+                  : null,
+              order: index + 1,
+            });
+          } else {
+            throw false;
+          }
         } else {
           transformed.album_artists.push({
             artist_id: aa.artist_id.id,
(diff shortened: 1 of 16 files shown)


robbevp:
@charvp rebased and updated with new eslint config

src/views/App.vue
```diff
@@ -200,36 +200,31 @@ export default {
...
+      if (this.isModerator) {
+        promises.push(this.$store.dispatch("rescan/show"));
+        promises.push(this.$store.dispatch("codecs/index"));
+        promises.push(this.$store.dispatch("codecConversions/index"));
+        promises.push(this.$store.dispatch("coverFilenames/index"));
+        promises.push(this.$store.dispatch("imageTypes/index"));
+        promises.push(this.$store.dispatch("locations/index"));
+      }
+      await Promise.all(promises);
+      this.loading = false;
```
charvp, line 223: This is a semantic change. If things go wrong in the above promises, `this.loading = false` will not be executed. Should be wrapped in a `try { ... } finally { ... }`.

src/views/genres/EditGenre.vue
```diff
@@ -60,14 +60,12 @@ export default {
...
-      this.update({ id: this.genre.id, newGenre: this.newGenre }).then(
-        (succeeded) => {
-          if (succeeded) {
-            this.$router.push(this.$route.query.redirect || { name: "genres" });
-          }
-        }
-      );
+    async submit() {
+      const newGenre = { id: this.genre.id, newGenre: this.newGenre };
+      const succeeded = await this.update(newGenre);
```
charvp, line 65: Why introduces the extra `const newGenre` here? I wouldn't do this; there are now three things called `newGenre` in this context.

src/views/labels/EditLabel.vue
```diff
@@ -60,14 +60,12 @@ export default {
...
-      this.update({ id: this.label.id, newLabel: this.newLabel }).then(
-        (succeeded) => {
-          if (succeeded) {
-            this.$router.push(this.$route.query.redirect || { name: "labels" });
-          }
-        }
-      );
+    async submit() {
+      const newLabel = { id: this.label.id, newLabel: this.newLabel };
+      const succeeded = await this.update(newLabel);
```
charvp, line 65: Same here.

src/components/AlbumForm.vue
```diff
@@ -375,52 +375,44 @@ export default {
         album_artists: [],
       };
 
-      const promises = [];
-
-      for (let label of this.newAlbum.album_labels) {
+      const label_map = this.newAlbum.album_labels.map(async (label) => {
```
charvp, line 378: There is also a small semantic change here (and this happens in almost every form/edit component). Previously, known labels would be added to the transformed object immediately. Now this will only happen when the `await Promise.all` happens below. I don't think this will actually result in a behavioural problem, so I guess it's fine.

charvp, line 378:
```suggestion
      const mappedLabels = this.newAlbum.album_labels.map(async (label) => {
``` 
Also note the use of `camelCase`; we only use `snake_case` for values in the store that come directly from the API. This should be fixed in a lot of places.

src/components/AlbumForm.vue
```diff
@@ -434,19 +426,17 @@ export default {
...
-          promise = this.create(transformed);
-        }
-        promise.then((succeeded) => {
-          if (succeeded) {
-            this.$router.push(this.$route.query.redirect || { name: "albums" });
-          }
-        });
-      });
+      await Promise.all([...artist_map, ...label_map]);
+      let promise = null;
```
charvp, line 430: `promise` feels like a bad name here. The whole point is that we're moving away from promises (should also be fixed in a lot of places).

src/components/Player.vue
```diff
@@ -238,20 +236,18 @@ export default {
...
-          .catch((error) => {
-            new Error(error);
-          });
+        try {
+          await this.$refs.audio.play();
+          if ("mediaSession" in navigator) {
+            navigator.mediaSession.playbackState = "playing";
+          }
+        } catch (error) {
+          new Error(error);
```
charvp, line 247: This doesn't make any sense. (I realise this was already something that didn't make any sense in the code that existed before, that doesn't mean it shouldn't be fixed.)

src/components/EditReviewComment.vue
```diff
@@ -53,12 +53,14 @@ export default {
     this.new_review_comment = this.item.review_comment;
```
charvp, line 53: Didn't notice this when the functionality was introduced, but `new_review_comment` should be camelCase.

charvp (changes requested):

src/components/Player.vue
robbevp: I've changed this to commit the error to the store, as we also do that when watching the `currentTrackURL`

src/components/AlbumForm.vue
robbevp: IMHO promise is still the correct term for a async function that is not yet resolved. `async/await` is another way of resolving promises.
But if you have another suggestion, I'll gladly change it

src/components/AlbumForm.vue
robbevp: I did test this, so the behaviour in the end will be the same.
I guess there could be a way to do it immediately, but i fear readability would suffer a lot

src/components/AlbumForm.vue
charvp: `pendingResult`?

src/components/AlbumForm.vue
charvp: The way to do it immediately would be the same as it was before: looping over it and adding it to an array in the case of an asynchronous operation.

charvp (approved):

PR #290, opened 2020-10-08
Add some sample tests with jest

robbevp:
I wanted to figure out how jest worked and try it on a few components. @charvp thoughts? Do we want to gradually build up some tests for the web interface?

For the setup, I mostly followed [the vuetify guide](https://v2.vuetifyjs.com/en/getting-started/unit-testing/) and [vue-test-utils docs](https://vue-test-utils.vuejs.org/guides/#getting-started)

Some thoughts:
It isn't always possible to find elements based on selectors. I've added some BEM-style classes to find them. This isn't ideal but it seemed better than depending on the order the elements where in.

We have a complex store, which makes that setting up a component for testing is quite complex. If we go further with this, we might want to create some factory's with fake data as we do in the API.


charvp:
I'm generally not a fan of this type of testing. We'll talk more about this in person.

PR #305, opened 2020-10-10
Fix linting errors

robbevp:
@charvp merging #276 after #301 caused some linting errors. Fixed here

### DIFF
--- a/src/store/genres.js
+++ b/src/store/genres.js
@@ -2,6 +2,7 @@ import Vue from "vue";
 import {
   index,
   create,
+  read,
   destroy,
   update,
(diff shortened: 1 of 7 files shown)


src/store/genres.js
```diff
@@ -2,6 +2,7 @@ import Vue from "vue";
 import {
   index,
   create,
+  read,
```
charvp, line 5: These changes are unrelated?

robbevp: No, was also something picked up by eslint.

charvp (approved):

Issue #303, opened 2020-10-10
Add issue and PR templates and write a CONTRIBUTING.md


charvp:

robbevp:
Closing since we have done both

charvp:
We don't have a PR template yet.

PR #309, opened 2020-10-13
Add basic issue templates

robbevp:
Adds a basic issue template. This is strongly based on github's default - I guess we'll update and improve as we go along?

re #303

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,41 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug_triage
+assignees: ''
+
+---
+
+<!--- Before you submit, make sure to check the following --->
+* [ ] There are no similar issues reported yet 
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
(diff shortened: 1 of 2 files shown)


robbevp:
We might want to create a separate one for questions? I don't think we can expect everyone to jump into the matrix channel. Or can we add the discussion tab like [vuetify](https://github.com/vuetifyjs/vuetify/discussions)?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,41 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug_triage
```
charvp, line 5: Just `bug`, I think.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: feature_request
```
charvp, line 5: We use `enhancement` in the accentor repos.

charvp (changes requested):

charvp:
> We might want to create a separate one for questions? I don't think we can expect everyone to jump into the matrix channel. Or can we add the discussion tab like [vuetify](https://github.com/vuetifyjs/vuetify/discussions)?

I don't like handling questions in GitHub issues. The matrix channel should be accessible to guests, so I don't think people even need to create an account. The discussions tab is a GitHub beta feature, so this can't be enabled (yet). Once it goes into public beta I don't have any qualms about adding it.

charvp (approved):

PR #319, opened 2020-10-17
Make albums sort more consistent

robbevp:
This PR does two things:
1. It provides a way to sort albums when `release` and/or `normalized_title` are equal.
In the current sorting, there were a few cases where we would still sort based on the orginal order (American Football's three s/t in when sorting by title, the two versions of Amenra's Mass VI when sorting by release or title).
If release and/or title don't result in an order, we look to the edition information (first the edition date and then the description). If that still doesn't provide an order, we look at the id to be sure we have a consistent order.

2. It  makes sorting albums more consistent.
in `compareTracks` we looked at `normalized_title`, `release` and `id`. If we sorted by title we usually only looked at the `normalized_title` and nothing else, while sorting by release also looked at `normalized_title`.
I've made two clear sorting orders (title first or release first) that both have enough fallback values so that we should always get the same order.

### DIFF
--- a/src/comparators.js
+++ b/src/comparators.js
@@ -10,7 +10,7 @@ export function compareStrings(s1, s2) {
 
 export function compareAlbumEditions(a1, a2) {
   // If one album has an edition but the other doesn't, we assume
-  // that the one without an edition is the "original" and schould come first.
+  // that the one without an edition is the "original" and should come first.
   if (a1.edition === null && a2.edition === null) {
     return 0;
   } else if (a1.edition === null) {
--- a/src/comparators.js
+++ b/src/comparators.js
@@ -38,18 +38,18 @@ export function compareTracks(rootState, t1, t2) {
     return -1;
   }
 
-  let albOrd = compareAlbumsByTitle(a1.normalized_title, a2.normalized_title);
+  let albOrd = compareAlbumsByTitleFirst(a1, a2);
   return albOrd === 0 ? t1.number - t2.number : albOrd;
 }
 
-export function compareAlbumsByTitle(a1, a2) {
+export function compareAlbumsByTitleFirst(a1, a2) {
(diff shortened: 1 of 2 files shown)


src/comparators.js
```diff
@@ -19,8 +38,23 @@ export function compareTracks(rootState, t1, t2) {
...
   }
 
-  let albOrd = compareStrings(a1.normalized_title, a2.normalized_title);
-  albOrd = albOrd === 0 ? compareStrings(a1.release, a2.release) : albOrd;
-  albOrd = albOrd === 0 ? compareStrings(a1.id, a2.id) : albOrd;
+  let albOrd = compareAlbumsByTitle(a1.normalized_title, a2.normalized_title);
   return albOrd === 0 ? t1.number - t2.number : albOrd;
 }
+
+export function compareAlbumsByTitle(a1, a2) {
```
charvp, line 45: These function names aren't accurate, IMO. It would be clearer to say `compareAlbumsByTitleFirst` or `compareForAlbumsSortedByTitle` (or something else that conveys that the albums are compared by more than their titles).

src/store/albums.js
```diff
@@ -165,38 +165,26 @@ export default {
   getters: {
     albums: (state) => Object.values(state.albums),
     albumsByTitle: (state, getters) =>
-      getters.albums.sort((a1, a2) =>
-        compareStrings(a1.normalized_title, a2.normalized_title)
-      ),
+      getters.albums.sort((a1, a2) => compareAlbumsByTitle(a1, a2)),
```
charvp, line 168:
```suggestion
      getters.albums.sort(compareAlbumsByTitle),
```
(Same for the others below.)

charvp (changes requested):

charvp (approved):

PR #365, opened 2020-11-30
Don't include undefined tracks in playlistTracks

### LINKED ISSUE #364: Player cannot be opened if track in playlist is deleted
- [X] There are no similar issues reported yet
Slightly related to #230 

**Describe the bug**
If a track inside the playlist is deleted, the Player can no longer be opened to reveal the playlist.
(Since we automatically skip tracks without audio or tracks that no longer exist, the audio will keep playing without problem)

**To Reproduce**

1. Add album to playlist
2. Delete track from album
3. Click on playlist toggle in player
4. Nothing happens

**Expected behavior**
The playlist should be visible without the deleted track

robbevp:
fix #364

### DIFF
--- a/src/store/player.js
+++ b/src/store/player.js
@@ -144,10 +144,9 @@ export default {
   },
   getters: {
     playlistTracks(state, getters, rootState) {
-      return state.playlist.reduce((acc, id) => {
-        const track = rootState.tracks.tracks[id];
-        return track ? [...acc, track] : acc;
-      }, []);
+      return state.playlist
+        .map((id) => rootState.tracks.tracks[id])
+        .filter((t) => t);
     },
     currentTrack(state, getters) {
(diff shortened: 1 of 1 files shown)


src/store/player.js
```diff
@@ -144,7 +144,10 @@ export default {
   },
   getters: {
     playlistTracks(state, getters, rootState) {
-      return state.playlist.map((id) => rootState.tracks.tracks[id]);
+      return state.playlist.reduce((acc, id) => {
+        const track = rootState.tracks.tracks[id];
+        return track ? [...acc, track] : acc;
+      }, []);
```
charvp, line 150: This creates a lot of new arrays unnecessarily. A filter + map would probably be better.

charvp (changes requested):

charvp (approved):

PR #379, opened 2020-12-12
Prevent rescan show from starting multiple loops

### LINKED ISSUE #371: Entering library settings always starts an extra rescan fetch
- [x] There are no similar issues reported yet

When opening the library settings page, a rescan fetch is performed, even when we are still in a previous rescan fetch loop.

**To Reproduce**
1. Start rescan
2. Switch a few times between home and library settings
3. Observe lots of requests in the network tab

**Expected behavior**
We should only have one rescan fetch loop.

robbevp:
## Description

Fix #371
Changes the rescan show action so that it only resolves once the rescan is no longer running.
If show is triggered a second time we return true immediately.

# How Has This Been Tested?

Tested locally, both when rescan is running and when it isn't

### DIFF
--- a/src/store/rescan.js
+++ b/src/store/rescan.js
@@ -39,6 +39,7 @@ export default {
         return true;
       } catch (error) {
         commit("addError", error, { root: true });
+        commit("setLoading", false);
         return false;
       }
     },
--- a/src/store/rescan.js
+++ b/src/store/rescan.js
@@ -1,30 +1,41 @@
 import { show, start } from "../api/rescan";
+import { wait } from "../utils";
 
 export default {
   namespaced: true,
   state: {
     rescan: null,
     lastClick: new Date(0),
+    loading: false,
   },
   mutations: {
     setLastClick(state, payload) {
       state.lastClick = payload;
+    },
+    setLoading(state, payload) {
+      state.loading = payload;
     },
     setRescan(state, payload) {
       state.rescan = payload;
     },
   },
   actions: {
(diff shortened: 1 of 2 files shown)


charvp (changes requested):
I don't really like this solution. It seems to me that `state.rescan.loading` is always the same as `          state.rescan.lastClick > new Date(result.finished_at) || result.running`? So adding that is a bit redundant. I would also not introduce `utils` until we actually need that functionality in multiple places (and hopefully that enlightens a better name than `utils`). I also don't see why you introduced a `do-while` loop? `lastClick` is updated before the `show` action, right? So the condition will always be true when the `try` is entered.

robbevp:
> I don't really like this solution. It seems to me that `state.rescan.loading` is always the same as ` state.rescan.lastClick > new Date(result.finished_at) || result.running`? So adding that is a bit redundant.

Since our `start` action sets the `lastClick` and sets `rescan.running` to true, it would block the show action from starting. We could solve this by not setting `running: true` on L49, but this could prove unreliable of the server already returns `running: true`.
I've a problem occurs while loading, a user might also get stuck not being able to load the new rescan state (if they started the rescan loop and the last result had `running: true`.

> I also don't see why you introduced a `do-while` loop? `lastClick` is updated before the `show` action, right? So the condition will always be true when the `try` is entered.

Because the show-loop isn't only started when a user triggered the rescan, but also we the first show returns a result as running. In the latter case we can only know whether to loop after we have received a response from the server.

charvp (approved):

PR #380, opened 2020-12-13
Improve reordering of TrackArtists

robbevp:
# Description

This PR brings two improvements to the way TrackArtists can be ordered.

## Drag and drop
I've added drag and drop support, so the user can quickly move an artist multiple steps (Especially relevent for tracks with many artists).
This only works through the drag-icon and is disabled if the length of TrackArtists is exactly one.

## Keypress controls
Currently the user has to tab multiple times to go past the controls (something which annoys me a lot).
To improve the UX if created one element around all the buttons with a few keypress events to control teh order.
A user can move the artist with ArrowUp, ArrowDown and U (for up) and D (for down).
A user can delete the artist through delete or backspace.

This does mean that there is no way to focus on the individual buttons. If this is necessary (but IMHO it isn't) we could allow the user to enter this level by pressing enter (and esc to leave).

(Do note that there is no way for a new user to know these things. I like the way how Hey gives this information, but this wouldn't fit our current UI.)

## Other small improvements
* Better spacing for order controls
* The EditTrack form is only shown after the track is loaded, to prevent a quick flash of the form in its empty state before the track is loaded 

# How Has This Been Tested?

Tested in chrome 87 and FF 83 on macOS, both for EditTrack page and MassEditDialog component

### DIFF
--- a/src/views/tracks/EditTrack.vue
+++ b/src/views/tracks/EditTrack.vue
@@ -9,7 +9,7 @@
         >
           {{ track.review_comment }}
         </VAlert>
-        <VForm v-model="isValid" @submit.prevent="submit">
+        <VForm v-model="isValid" @submit.prevent="submit" v-if="loaded">
           <VTextField
             type="number"
             :label="$t('music.track.number')"
@@ -104,17 +104,20 @@ export default {
       clear_review_comment: true,
       isDirty: false,
       isValid: true,
+      loaded: false,
     };
   },
   async created() {
     if (this.track) {
       await this.read(this.track.id);
+      this.loaded = true;
       this.fillValues();
     }
   },
   watch: {
     track: function () {
       if (this.track && !this.isDirty) {
+        this.loaded = true;
         this.fillValues();
       }
     },
--- a/src/components/TrackFormArtists.vue
+++ b/src/components/TrackFormArtists.vue
@@ -11,7 +11,7 @@
           class="d-flex justify-space-between fill-height flex-column py-2"
         >
           <VBtn
-            @click="moveArtist({ oldIndex: index, newIndex: index - 1 })"
+            @click="moveArtist(index, -1)"
             icon
             small
             class="ma-2"
@@ -31,7 +31,7 @@
             <VIcon>mdi-drag-horizontal-variant</VIcon>
           </VBtn>
           <VBtn
-            @click="moveArtist({ oldIndex: index, newIndex: index + 1 })"
+            @click="moveArtist(index, 1)"
             icon
             small
             class="ma-2"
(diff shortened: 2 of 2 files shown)


charvp:
> # Description
> 
> This PR brings two improvements to the way TrackArtists can be ordered.
> ## Drag and drop
> 
> I've added drag and drop support, so the user can quickly move an artist multiple steps (Especially relevent for tracks with many artists).
> This only works through the drag-icon and is disabled if the length of TrackArtists is exactly one.
> ## Keypress controls
> 
> Currently the user has to tab multiple times to go past the controls (something which annoys me a lot).
> To improve the UX if created one element around all the buttons with a few keypress events to control teh order.
> A user can move the artist with ArrowUp, ArrowDown and U (for up) and D (for down).
> A user can delete the artist through delete or backspace.
> 

> 
> (Do note that there is no way for a new user to know these things. I like the way how Hey gives this information, but this wouldn't fit our current UI.)
> ## Other small improvements
> 
>     * Better spacing for order controls
> 
>     * The EditTrack form is only shown after the track is loaded, to prevent a quick flash of the form in its empty state before the track is loaded
> 
> 
> # How Has This Been Tested?
> 
> Tested in chrome 87 and FF 83 on macOS, both for EditTrack page and MassEditDialog component

charvp:
Sorry, wrong button.

charvp (changes requested):
> This does mean that there is no way to focus on the individual buttons. If this is necessary (but IMHO it isn't) we could allow the user to enter this level by pressing enter (and esc to leave).

I like to edit with only keyboard controls, so this does seem like a requirement for me. To be honest, I don't see the use for the extra keyboard controls. The drag and drop is nice though.

robbevp:
> > This does mean that there is no way to focus on the individual buttons. If this is necessary (but IMHO it isn't) we could allow the user to enter this level by pressing enter (and esc to leave).
> 
> I like to edit with only keyboard controls, so this does seem like a requirement for me. To be honest, I don't see the use for the extra keyboard controls. The drag and drop is nice though.

Have you tried editing a few tracks with the keyboard controls? There might be some adjustment time, but I found that the new keyboard controls allow me to edit faster since (1) I don't have to tab through up, down, delete for each artist (2) it's a lot easier to move an artist multiple places since the keyboard focus follows the artist you are moving.
If you disagree, I'll add a way to focus on the individual buttons

src/views/tracks/EditTrack.vue
```diff
@@ -12,7 +12,7 @@
         >
           {{ track.review_comment }}
         </VAlert>
-        <VForm v-model="isValid" @submit.prevent="submit">
+        <VForm v-model="isValid" @submit.prevent="submit" v-if="loaded">
```
chvp, line 15: Try to refrain from fixing multiple issues in one pull request in the future.

chvp (approved):

chvp (approved):

PR #393, opened 2021-01-04
Fix order of artists through mass edit

### LINKED ISSUE #34: Newly created artists aren't added in the right order on mass edit

robbevp:
## Description

Fix #34 
We previously set the order of the artists in a track based on the length, which changed while we were adding the artists.
This would mess up the order if for some reason the transformedArtists array is not in the same order as the order property. (Which seems to happen with new artists)

We now set the order based on the length of the length before we start adding.
I've dropped the `-1` since this would generate orders like `1, 1, 2, 3`.

# How Has This Been Tested?

Tested locally with a number of tracks

### DIFF
--- a/src/components/MassEditDialog.vue
+++ b/src/components/MassEditDialog.vue
@@ -491,7 +491,7 @@ export default {
                     ta.artist_id === a.artist_id
                 ).length === 0
               ) {
-                a.order += t.track_artists.length - 1;
+                a.order += t.track_artists.length;
                 transformed.track_artists.push(a);
               }
             });
--- a/src/components/MassEditDialog.vue
+++ b/src/components/MassEditDialog.vue
@@ -457,7 +457,7 @@ export default {
           album_id: t.album_id,
           review_comment: this.clearReviewComments ? null : t.review_comment,
           genre_ids: t.genre_ids,
-          track_artists: t.track_artists,
+          track_artists: [...t.track_artists],
         };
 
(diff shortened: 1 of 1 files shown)


charvp (changes requested):
This fix doesn't seem right to me. Changes to the `transformed` track shouldn't reflect on the original `t` track. Or am I missing something?

robbevp:
> This fix doesn't seem right to me. Changes to the `transformed` track shouldn't reflect on the original `t` track. Or am I missing something?

You're right. Apparently the problem started when we create `transformed`: both `t.track_artists` and `transformed.track_artists` referred to the same array, so when we pushed to `transformed.track_artists` we also mutated `t.track_artists`. Fixed by destructuring and creating a new array

src/components/MassEditDialog.vue
```diff
@@ -457,7 +457,7 @@ export default {
           album_id: t.album_id,
           review_comment: this.clearReviewComments ? null : t.review_comment,
           genre_ids: t.genre_ids,
```
charvp, line 459: I guess this should also happen for `genre_ids`?

charvp (changes requested):
Doesn't the -1 thing still apply as well?

src/components/MassEditDialog.vue
robbevp: true, fixed

charvp (approved):

PR #395, opened 2021-01-05
Fix spacing with VRow

### LINKED ISSUE #394: Wrong vertical spacing on various pages/components
- [x] There are no similar issues reported yet

**Describe the bug**
Probably due to a change in vuetify (though they mention no breaking changes in the last few release notes).
Effected pages:
- [ ] Library
- [ ] Player

robbevp:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fix #394 

This negates the changes introduced by https://github.com/vuetifyjs/vuetify/commit/d0f25fc59b29c385a6910dc48a111df811da3bdf#diff-40030629d72eca85d1eda4223b61cf497956726ca57bab2e64813010baf80cd8
The fix is done by
* Adding a `no-gutters` prop, which sets all margins to 0
* Manually setting vertical or bottom margins.
* Wrapping the content of the row in `VCol` component

## How Has This Been Tested?

* I've tried to compare with our app before vuetify 2.4.0. 
* I went through all occurences of `VRow` to see where we needed to fix the margins

### DIFF
--- a/src/components/AlbumForm.vue
+++ b/src/components/AlbumForm.vue
@@ -201,7 +201,7 @@
           v-model="clear_review_comment"
           :label="$tc('music.flag.clear', 1)"
         />
-        <VRow justify="center" no-gutters>
+        <VRow justify="center my-o">
           <VBtn :disabled="!isValid" color="primary" class="ma-2" type="submit">
             {{
               this.album ? $t("music.album.update") : $t("music.album.create")
--- a/src/components/AlbumForm.vue
+++ b/src/components/AlbumForm.vue
@@ -201,7 +201,7 @@
           v-model="clear_review_comment"
           :label="$tc('music.flag.clear', 1)"
         />
-        <VRow justify="center">
+        <VRow justify="center" no-gutters>
           <VBtn :disabled="!isValid" color="primary" class="ma-2" type="submit">
             {{
               this.album ? $t("music.album.update") : $t("music.album.create")
--- a/src/components/AlbumForm.vue
+++ b/src/components/AlbumForm.vue
@@ -201,7 +201,7 @@
           v-model="clear_review_comment"
           :label="$tc('music.flag.clear', 1)"
         />
-        <VRow justify="center my-o">
+        <VRow justify="center my-0">
           <VBtn :disabled="!isValid" color="primary" class="ma-2" type="submit">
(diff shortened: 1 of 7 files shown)


charvp:
Seems like this is still being worked on on the vuetify side: https://github.com/vuetifyjs/vuetify/issues/12915 https://github.com/vuetifyjs/vuetify/pull/12916
Are we running into the same thing? If so, I would just wait on the vuetify fix. (Adding `no-gutters` doesn't really make sense to add spacing IMO.)

charvp (commented):
Seems like this is still being worked on on the vuetify side: https://github.com/vuetifyjs/vuetify/issues/12915 https://github.com/vuetifyjs/vuetify/pull/12916
Are we running into the same thing? If so, I would just wait on the vuetify fix. (Adding `no-gutters` doesn't really make sense to add spacing IMO.)

robbevp:
> Seems like this is still being worked on on the vuetify side: [vuetifyjs/vuetify#12915](https://github.com/vuetifyjs/vuetify/issues/12915) [vuetifyjs/vuetify#12916](https://github.com/vuetifyjs/vuetify/pull/12916)
> Are we running into the same thing? If so, I would just wait on the vuetify fix. (Adding `no-gutters` doesn't really make sense to add spacing IMO.)

I don't think that issue applies, since it is explicetly about `VRow dense`, which we don't use.
If you look at https://github.com/vuetifyjs/vuetify/issues/12848 - it seems to me that Vuetify doesn't think this is an issue but rather intended. I would not wait on the off chance they change this behaviour. 

`no-gutters` to fix can be confusing in the future. Changed to a `my-0` utility class

src/components/AlbumForm.vue
```diff
@@ -201,7 +201,7 @@
           v-model="clear_review_comment"
           :label="$tc('music.flag.clear', 1)"
         />
-        <VRow justify="center">
+        <VRow justify="center my-o">
```
charvp, line 204: `o`?

charvp (changes requested):
I still don't really get how this can be the intended behaviour by vuetify 🤷‍♀️, but I guess they're pretty set on it.

charvp (approved):

PR #406, opened 2021-01-10
Allow user to sort TrackTable

robbevp:
# Description

Allows the user to sort the TrackTable for all relevant properties by providing the `VDataTable` with a method to sort all properties. If the user doesn't determine the sort, the items are sorted as provided to the component.

* `compareTracks` takes an album array as an optional param. I had first reworked this to take an array of albums rather than the rootState, but it makes more sense to not require any array.
* The `TrackTable` has one method to sort all items. (Vuetify also allows a `sort` prop in the header, but this only has access to that prop, which doesn't work for us. We often need the whole track.) This method only uses the first item of `sortBy` and `sortDesc` since I have not enabled `multiSort`
* Added a function to sort tracks by genres or artists.
* Headers all have a class `text-no-wrap`. Without this the icon to sort would often appear underneath or above the header, which make for a very ugly layout.

# How Has This Been Tested?

Extensive local tests to make sure properties are sorted as expected.

### DIFF
--- a/src/comparators.js
+++ b/src/comparators.js
@@ -74,48 +74,46 @@ export function compareTracksByArtist(t1, t2) {
   return order;
 }
 
-export function compareTracksByGenre(
-  t1,
-  t2,
-  genres = store.state.genres.genres
-) {
-  // tracks without genres get sorted last
-  if (t1.genre_ids.length === 0 && t2.genre_ids.length === 0) {
-    return 0;
-  } else if (t1.genre_ids.length === 0) {
-    return 1;
-  } else if (t2.genre_ids.length === 0) {
-    return -1;
-  }
-
-  // we map the genres for each track and sort by name
-  const t1_genres = t1.genre_ids
-    .map((g) => genres[g].normalized_name)
-    .sort((g1, g2) => compareStrings(g1, g2));
-  const t2_genres = t2.genre_ids
-    .map((g) => genres[g].normalized_name)
-    .sort((g1, g2) => compareStrings(g1, g2));
-
-  // We loop over the sorted genres until we find a difference to sort by
-  let index = 0;
-  let order = 0;
-  while (order === 0) {
-    const g1 = t1_genres[index];
-    const g2 = t2_genres[index];
-
-    // If tracks share the same genres, but one track has one or more extra genre(s),
-    // we sort the track with more genres last
-    if (typeof g1 === "undefined" && typeof g2 === "undefined") {
+export function compareTracksByGenre(genres) {
+  return function (t1, t2) {
+    // tracks without genres get sorted last
+    if (t1.genre_ids.length === 0 && t2.genre_ids.length === 0) {
       return 0;
-    } else if (typeof g1 === "undefined") {
-      return -1;
-    } else if (typeof g2 === "undefined") {
+    } else if (t1.genre_ids.length === 0) {
       return 1;
+    } else if (t2.genre_ids.length === 0) {
+      return -1;
     }
-    order = compareStrings(g1, g2);
-    index++;
-  }
-  return order;
+
+    // we map the genres for each track and sort by name
+    const t1_genres = t1.genre_ids
+      .map((g) => genres[g].normalized_name)
+      .sort((g1, g2) => compareStrings(g1, g2));
+    const t2_genres = t2.genre_ids
+      .map((g) => genres[g].normalized_name)
+      .sort((g1, g2) => compareStrings(g1, g2));
+
+    // We loop over the sorted genres until we find a difference to sort by
+    let index = 0;
+    let order = 0;
+    while (order === 0) {
+      const g1 = t1_genres[index];
+      const g2 = t2_genres[index];
+
+      // If tracks share the same genres, but one track has one or more extra genre(s),
+      // we sort the track with more genres last
+      if (typeof g1 === "undefined" && typeof g2 === "undefined") {
+        return 0;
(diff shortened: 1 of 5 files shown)


src/comparators.js
```diff
@@ -1,3 +1,5 @@
+import store from "@/store/store";
```
chvp, line 1: Importing the store here doesn't feel very nice, this should be a utility file. What is wrong with always passing the albums as a param?

src/components/TracksTable.vue
```diff
@@ -77,7 +78,7 @@
       </template>
       <template v-slot:item.album_id="props">
         <RouterLink :to="{ name: 'album', params: { id: props.value } }">
-          {{ albums[props.value].title }}
+          {{ albums[props.value] ? albums[props.value].title : "" }}
```
chvp, line 81: Is this not always present?

src/components/TracksTable.vue
```diff
@@ -202,6 +210,31 @@ export default {
...
+        case "title":
+          sortFunction = (t1, t2) =>
+            compareStrings(t1.normalized_title, t2.normalized_title);
+          break;
+        case "track_artists":
+          sortFunction = compareTracksByArtist;
+      }
+      const sorted = sortFunction ? items.sort(sortFunction) : items;
+      return sortDesc[0] ? sorted.reverse() : sorted;
+    },
```
chvp, line 237: I'm not sure about all these options. In general, I would prefer to only add features we have a clear use case for. Sorting by number for example doesn't seem like it would ever be useful.

chvp (changes requested):
Following on from my last comment, I'm not entirely sure this is how we want to implement sorting. For example, one useful sorting order that is still impossible is to sort the albums by year and keep the order within albums the same. This is currently impossible with this implementation. Some kind of dropdown would allow us more freedom.

src/components/TracksTable.vue
robbevp: No, it could always be that a track is loaded before the album is. But I'll extract this to a separate PR, since this is a bug fix

robbevp:
> Following on from my last comment, I'm not entirely sure this is how we want to implement sorting. For example, one useful sorting order that is still impossible is to sort the albums by year and keep the order within albums the same. This is currently impossible with this implementation. Some kind of dropdown would allow us more freedom.

I don't really follow. If a user wants to sort by year, the year should be visible inside the table, no? Sorting by something you can't see seems like bad UX to me.
Since we don't allow multi-param sorting, the order after the selected param is up to us (as we already do for sorting by album title)

src/components/TracksTable.vue
robbevp: We sort by number in the album view.
Moreover, if we keep sorting inside the table headers I wouldn't restrict these because we don't see the use. It feels weridly restrictive to only allow certain properties. If a user wants to sort all their tracks by number (because they want to know which tracksnr in their library is the highest or whatever) they should be able to

src/comparators.js
robbevp: This was mostly to avoid that we have to write `sort((t1, t2) -> compareTracks(t1, t2, rootState))` and to make the sorting function unaware of the structure of our store.
I've rewritten this so that we don't have to import the store here.

chvp:
I get where you're coming from, but sorting albums by year in the track table is one of the things that I would like to see the most (especially on the artist view). A lot more than sorting by track length or number, in any case. I don't really see how it would be bad UX if the date isn't shown in the table, users know that albums have release dates.

Code changes look good :+1:

robbevp:
> I get where you're coming from, but sorting albums by year in the track table is one of the things that I would like to see the most (especially on the artist view). A lot more than sorting by track length or number, in any case. I don't really see how it would be bad UX if the date isn't shown in the table, users know that albums have release dates.


I agree that a sorting albums by year is relevant. A user will indeed know that an album has a year/release date, but they won't know what that year is.

Since data might be wrong or less specific (ie when only the year is known instead of the exact date) a user might expect a different result then shown. Or they might not know a discography and want to know more. There wouldn't be an immediate way to know way the tracks were sorted in that way (only clicking on an album, going back, clicking on another album, ...). That seems like bad UX to me. If a user can sort by a property, they should be able to see that property.

chvp:
I don't really agree, the album dates are usually right above (on the artist view). However, that isn't the main point to me. When adding features, I am of the opinion that the use case for the feature is the most important aspect. If we start from use cases we want to support, you seem to agree that sorting albums by date in the track table is something we want to support (to me, it's the thing I want to be able to do the most). The way sorting is implemented here, that would make it a harder feature to support instead of an easier one.

robbevp:
I disagree with both of your points.
* I would only want to sort by year in the track table if we also show the year in the table. Have a sorting function that only makes sense in one particular view and then still requires the user to scroll up and down is exactly what I mean with bad UX. 
* The way sorting is currently implemented would not make this harder. If we'd ever change this UI (which again, I wouldn't be in favor of), we'll just need to set `sortBy` and `sortDesc` as props on the `VDataTable` instead of the component handling this selection for us. The current function to sort all the items easily allows to add (or remove) `sortBy` properties.

chvp:
https://github.com/accentor/android/issues/280 kind of swayed me to your viewpoint. Can you rebase? Then I'll look at the code once again and merge if it seems fine.

robbevp:
@chvp rebased

chvp (approved):

PR #413, opened 2021-01-16
(Re)load content on main objects pages

robbevp:
# Description
This PR adds all functionality to (re)load the content on the main object pages (Album, Artist, Genre, Label).
This means:
* Going directly to a specifc page (`app/albums/1`) will get content and functionality almost instantly.
* Content is much more likely to stay up to date.

There are two drawbacks to this:
* We introduce a lot more API calls. Every page for a main object will do 2 -3 API calls. I would try this and see if it creates issues/too much traffic. If so, we can check `object.loaded` before calling the api and only call the API if the content is older than some timeperiod.
* There is a higher chance that related content is missing. This is something that should/will be added in a future PR.

## Scopes
To accomplish this, I've added a few scopes classes to our api interface.
These classes allow the user to easily create a scope without bothering with the implementation on the api-side.
A scope can be:
* created on a single line `new AlbumsScope.label(id)`
* chained for more complex queries `new AlbumsScope.label(id).artist(id).filter(string)`
* created and then modified:
```
const scope = new Albumscope()
scope.label(id)
scope.artist(id)
scope.finalQuery
```

The scope class throws errors if we pass it the wrong kind of parameter. This errors are not translated since the end-user should not encounter these. They are only their to notify the develop if they made an error when building scopes.

If we use a scope, we skip `removeOld` after fetching all data. In the future, we might want to allow that method check whether any objects that fit the scope should be removed, but IMHO that is not really relevant. (If a user visits the page for an object that doesn't exist on the api anymore, they will get an error and go back to the previous page).

# How Has This Been Tested?

- [x] Local tests to see if everything works
- [x] A quick test with our production api, as a final check.

### DIFF
--- a/src/api/fetch.js
+++ b/src/api/fetch.js
@@ -6,19 +6,22 @@ const fetchRetry = require("fetch-retry")(fetch, {
   },
 });
 
-export async function* indexGenerator(path, auth) {
+export async function* indexGenerator(path, auth, scope = { finalQuery: "" }) {
   let page = 1;
   while (true) {
     let response, result;
     try {
-      response = await fetchRetry(`${baseURL}/${path}?page=${page}`, {
-        retries: 5,
-        method: "GET",
-        headers: {
-          "x-secret": auth.secret,
-          "x-device-id": auth.device_id,
-        },
-      });
+      response = await fetchRetry(
+        `${baseURL}/${path}?page=${page}${scope.finalQuery}`,
+        {
+          retries: 5,
+          method: "GET",
+          headers: {
+            "x-secret": auth.secret,
+            "x-device-id": auth.device_id,
+          },
+        }
+      );
       result = await response.json();
     } catch (reason) {
       throw { error: [reason] };
--- a/src/api/fetch.js
+++ b/src/api/fetch.js
@@ -1,4 +1,5 @@
 import baseURL from "./base_url";
+import { Scope } from "./scopes";
 const fetchRetry = require("fetch-retry")(fetch, {
   retries: 0,
   retryDelay: function (attempt) {
@@ -6,7 +7,7 @@ const fetchRetry = require("fetch-retry")(fetch, {
   },
 });
 
-export async function* indexGenerator(path, auth, scope = { finalQuery: "" }) {
+export async function* indexGenerator(path, auth, scope = new Scope()) {
   let page = 1;
   while (true) {
     let response, result;
--- a/src/api/scopes.js
+++ b/src/api/scopes.js
@@ -5,52 +5,55 @@ export class Scope {
     return this.scopes.map((s) => `&${s.key}=${s.query}`).join("");
   }
 
-  addScope = (key, query) => {
+  addScope(key, query) {
     this.scopes.push({ key, query });
     return this;
-  };
+  }
 
-  addScopesFromArray = (key, queries) => {
+  addScopesFromArray(key, queries) {
     queries.forEach((q) => this.addScope(key, q));
     return this;
-  };
+  }
 }
 
 export class AlbumsScope extends Scope {
-  artist = (id) => {
+  artist(id) {
     return this.addScope("artist_id", id);
-  };
+  }
 
-  filter = (string) => {
+  filter(string) {
     return this.addScope("filter", string);
-  };
+  }
 
-  label = (id) => {
+  label(id) {
     return this.addScope("label", id);
-  };
+  }
 
-  labels = (ids) => {
+  labels(ids) {
     return this.addScopesFromArray("labels", ids);
-  };
+  }
 }
 
 export class ArtistsScope extends Scope {
-  filter = (string) => {
+  filter(string) {
     return this.addScope("filter", string);
-  };
+  }
 }
 
 export class TracksScope extends Scope {
-  album = (id) => {
+  album(id) {
     return this.addScope("album_id", id);
-  };
-  artist = (id) => {
+  }
+
+  artist(id) {
     return this.addScope("artist_id", id);
-  };
-  filter = (string) => {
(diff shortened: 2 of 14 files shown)


src/api/fetch.js
```diff
@@ -1,24 +1,28 @@
...
+import { Scope } from "./scopes";
 const fetchRetry = require("fetch-retry")(fetch, {
   retries: 0,
   retryDelay: function (attempt) {
     return Math.pow(2, attempt) * 15000;
   },
 });
 
-export async function* indexGenerator(path, auth) {
+export async function* indexGenerator(path, auth, scope = new Scope()) {
```
chvp, line 10: From an API design standpoint, the optionality of this parameter should be clear in the `index` methods, not here.

src/api/scopes.js
```diff
@@ -0,0 +1,72 @@
+export class Scope {
+  scopes = [];
+
+  get finalQuery() {
+    return this.scopes.map((s) => `&${s.key}=${s.query}`).join("");
+  }
+
+  addScope = (key, query) => {
```
chvp, line 8: Is there a reason you don't write
```suggestion
  addScope(key, query) {
```
?

src/api/scopes.js
```diff
@@ -0,0 +1,72 @@
...
+    if (typeof query !== "string" && typeof query !== "number") {
```
chvp, line 9: This should not be necessary.

src/api/scopes.js
```diff
@@ -0,0 +1,72 @@
...
+      throw TypeError(
+        `${query} is an ${typeof query} while your scope '${key}' that expects a string or number.`
+      );
+    }
+    this.scopes.push({ key, query });
+    return this;
+  };
+
+  addScopesFromArray = (key, queries) => {
+    if (!Array.isArray(queries)) {
```
chvp, line 19: Same here.

src/store/actions.js
```diff
@@ -13,5 +13,8 @@ export async function fetchAll(commit, generator, commitAction) {
     }
   }
   commit(commitAction, results);
-  commit("removeOld");
+  // If a scope is present, we skip removeOld
+  if (scope === null) {
```
chvp, line 17: This does not do what the comment says it does.

src/views/albums/Album.vue
```diff
@@ -61,27 +61,33 @@
...
 
 export default {
   name: "Album",
   components: { AlbumArtists, TracksTable, AlbumActions },
   metaInfo() {
     return { title: this.album?.title };
   },
+  props: {
+    id: {
+      type: [String, Number],
```
chvp, line 78: The `[String, Number]` feels weird to me. Will this not always be a `String` since it is loaded from the URL? Or does `vue-router` only read the prop from the URL if it was not passed in code?

src/views/albums/Album.vue
```diff
@@ -102,6 +108,24 @@ export default {
...
+    ...mapActions("tracks", { indexTracks: "index" }),
+    async fetchContent(newValue, oldValue) {
+      // After loading the content, the router will change the id from a string to a number
+      // but we don't actually want to load the content twice
+      if (newValue == oldValue) {
+        return;
+      }
+
+      const album = this.read(this.id);
+      const tracks = this.indexTracks(new TracksScope().album(this.id));
```
chvp, line 121: It feels a little arbitrary what is fetched here. Why not fetch the genres of the tracks, why not fetch the labels?

src/views/albums/Album.vue
```diff
@@ -102,6 +108,24 @@ export default {
...
+      Promise.all([album, tracks]).finally(() => {
```
chvp, line 122: Why not `return await` this, to make the control flow in the async method explicit?

chvp (changes requested):

src/api/fetch.js
robbevp: I've also made it clear in the index action where applicable. We also need the optionality here, since not all index actions will provide a scope

src/views/albums/Album.vue
robbevp: The latter. If you go to `app/albums/1` directly the prop will be a string (and Vue's type checker will complain if that is not an option), if you navigate through the app the prop will be a number.

src/api/scopes.js
robbevp: Not really, it shouldn't make any difference to write it one way or the other

src/views/albums/Album.vue
robbevp: I mostly started from the scopes that are available. There isn't (yet) a scope to get genres by tracks or labels by album on the API.

src/api/scopes.js
chvp: I like the `function() {}` syntax more, tbh.

src/views/albums/Album.vue
chvp: OK.

src/views/albums/Album.vue
chvp: Ah, makes sense.

src/views/albums/Album.vue
```diff
@@ -61,27 +61,33 @@
...
 
 export default {
   name: "Album",
   components: { AlbumArtists, TracksTable, AlbumActions },
   metaInfo() {
     return { title: this.album?.title };
   },
+  props: {
+    id: {
+      type: [String],
```
chvp, line 78: Why did you remove the `, Number` here?

robbevp: I was experimenting with these types to reconstruct why I added both. This slipped in by accident.

chvp (approved):

PR #414, opened 2021-01-18
Only set localStorage if one of stored modules changes

robbevp:
# Description
Only set localStorage if one of stored modules changes

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Quick scan in localStorage to see if data that should be stored is present

PR #415, opened 2021-01-19
Cache vuex store

### LINKED ISSUE #39: Cache DB in browser

robbevp:
# Description
Fix #39
This uses the same idea as #48 but with a few differences.

## Different caches
We store data in two places: localStorage and localForage. localStorage is prefered since it is synchronous, so it is used for most modules. This way we have **instant** access to data from those modules

The main collections that are to big to be stored inside localStorage and are stored in localForage.
I've added a subclass to `VuexPersistence` that allows us to save and restore data in a memory safe way. (VuexPersist would try and merge the existing store with the savedStore, which would cause huge spikes in memory usage), we can now use the mutations within each module.

## Router isn't blocked
I've placed the `await store.restored` before we would commit something in any of the modules that are stored async. Since we overwrite the way this value is set, we have a separate restored for each module that is stored async.

This way we can already render the page while data is loaded from localForage. A user might experience a 2-3 sec delay before the page is populated with data, but that feels very acceptable.

## Filter mutations
Just as in #414. We only set a cache when something inside of the modules in the cache is changed.
This makes that commits in the player don't have any effect on the cache.

## Debounced `saveState`
The methods that are stored in localForage have their `saveState` wrapped in a debounce with a wait time of a minute. 
If a user is editing, loading data, it doesn't make much sense to save the state after every mutation. Since we don't use these values directly, the user should not notice this. 
(There is a slim chance that the user closes the tab before the data was stored, but if that is the case, the user will get the updated data soon after loading the app.)

# How Has This Been Tested?

Tested by refreshing quite a bit during the course of a day in both Chrome and FF.

The usual memory usage might spike up to 200MB when restoring or saveing the cache from localForage and settled to around 80MB - 120MB after this.

### DIFF
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -13,6 +13,7 @@ import {
   compareAlbumsByTitleFirst,
 } from "../comparators";
 import { AlbumsScope } from "../api/scopes";
+import store from "./store";
 
 export default {
   namespaced: true,
@@ -108,6 +109,7 @@ export default {
     async index({ commit, rootState }, scope = new AlbumsScope()) {
       const generator = index(rootState.auth, scope);
       try {
+        await store.restored;
         await fetchAll(commit, generator, "setAlbums", scope);
         return true;
       } catch (error) {
@@ -128,6 +130,7 @@ export default {
     async read({ commit, rootState }, id) {
       try {
         const result = await read(rootState.auth, id);
+        await store.restored;
         commit("setAlbum", { id, album: result });
         return result.id;
       } catch (error) {
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -109,7 +109,7 @@ export default {
     async index({ commit, rootState }, scope = new AlbumsScope()) {
       const generator = index(rootState.auth, scope);
       try {
-        await store.restored;
+        await store.albumsRestored;
         await fetchAll(commit, generator, "setAlbums", scope);
         return true;
       } catch (error) {
@@ -130,7 +130,7 @@ export default {
     async read({ commit, rootState }, id) {
       try {
         const result = await read(rootState.auth, id);
-        await store.restored;
+        await store.albumsRestored;
         commit("setAlbum", { id, album: result });
         return result.id;
       } catch (error) {
--- a/src/store/albums.js
+++ b/src/store/albums.js
@@ -13,7 +13,6 @@ import {
   compareAlbumsByTitleFirst,
 } from "../comparators";
 import { AlbumsScope } from "../api/scopes";
-import store from "./store";
 
 export default {
   namespaced: true,
@@ -109,7 +108,7 @@ export default {
     async index({ commit, rootState }, scope = new AlbumsScope()) {
       const generator = index(rootState.auth, scope);
       try {
-        await store.albumsRestored;
+        await this.albumsRestored;
         await fetchAll(commit, generator, "setAlbums", scope);
         return true;
       } catch (error) {
@@ -130,7 +129,7 @@ export default {
     async read({ commit, rootState }, id) {
       try {
         const result = await read(rootState.auth, id);
-        await store.albumsRestored;
+        await this.albumsRestored;
         commit("setAlbum", { id, album: result });
         return result.id;
       } catch (error) {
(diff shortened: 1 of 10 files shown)


robbevp:
I've also created a PR with vuex-persist: https://github.com/championswimmer/vuex-persist/pull/218
With this change we can cache tracks without too much of a performance hit

src/store/albums.js
```diff
@@ -12,6 +12,7 @@ import {
   compareAlbumsByReleaseFirst,
   compareAlbumsByTitleFirst,
 } from "../comparators";
+import store from "./store";
```
chvp, line 15: Importing the store from a store submodule doesn't feel very nice. Is there no way around this?

src/store/persistence.js
```diff
@@ -0,0 +1,96 @@
+import VuexPersistence from "vuex-persist";
+import localForage from "localforage";
+import debounce from "lodash.debounce";
+
+// A subclass for VuexPersintence to deal with single namepaced modules and large collections
```
chvp, line 5:
```suggestion
// A subclass for VuexPersistence to deal with single namespaced modules and large collections
```

src/store/persistence.js
```diff
@@ -0,0 +1,96 @@
...
+    // Get item from an async storage and commit them back to the store
+    this.replaceState = async (module, storage, store) => {
+      const savedState = await storage.getItem(module);
+      if (savedState) {
+        if (savedState.startLoading) {
+          store.commit(`${module}/setStartLoading`, savedState.startLoading);
+        }
+        if (savedState[module]) {
+          store.commit(
+            `${module}/set${module.charAt(0).toUpperCase() + module.slice(1)}`,
```
chvp, line 39: Don't really like the manual capitalize. Can we not pass this in the constructor?

chvp (changes requested):

src/store/albums.js
robbevp: Not that I'm aware of. I made some attempts to have the `restored` property as part of the module, or kept inside the store itself, but didn't find a way to make it work

src/store/persistence.js
robbevp: I'm not sure I follow. Do you mean to to have a separate argument with the module name capitalized? That seems redundant to me (or we would just do this function inside the map where we create the instances).
Two other options would be:
* To extract this to a separate function, so we can just call `capitalize(module)`
* Pass the constructor a capitalized module name and lowercase it for all other places where we need it

chvp: I'd prefer the second option then.

src/store/albums.js
chvp: OK :(

src/store/persistence.js
```diff
@@ -0,0 +1,114 @@
...
+      storage,
+      asyncStorage: true,
+      modules: [lowerCaseModule],
+      key: lowerCaseModule,
+      filter: (m) =>
+        m.type.substring(0, m.type.indexOf("/")) === lowerCaseModule,
+      saveState: debounce(saveModule, 60000),
+    });
+
+    this.upperCaseModule = module;
```
chvp, line 32:
```suggestion
    this.capitalizedModule = module;
```

chvp (changes requested):
Seems more correct to me. Feel free to disagree.

src/store/persistence.js
robbevp: I agree, changed

src/store/persistence.js
```diff
@@ -0,0 +1,114 @@
+import VuexPersistence from "vuex-persist";
+import localForage from "localforage";
+import debounce from "lodash.debounce";
+
+// A subclass for VuexPersintence to deal with single namespaced modules and large collections
```
chvp, line 5:
```suggestion
// A subclass for VuexPersistence to deal with single namespaced modules and large collections
```

chvp (dismissed):

chvp (changes requested):
Sorry, didn't see the merge conflicts before approving.

robbevp:
@chvp rebased

chvp (approved):

PR #525, opened 2021-07-08
Update license

robbevp:
Updates year in license


chvp (changes requested):
I'd prefer to just change the year to the current year instead of providing a range (the year in a license doesn't mean that much anyway, and the project was started in 2019).

chvp (approved):

PR #529, opened 2021-07-11
Display play count in TracksTable

robbevp:
Adds a very simple play count to the TracksTable.
I wasn't entirly sure where to place it, alternative suggestions are welcome.

### DIFF
--- a/src/components/TracksTable.vue
+++ b/src/components/TracksTable.vue
@@ -83,11 +83,11 @@
       <template v-slot:item.track_artists="props">
         <TrackArtists :track="props.item" />
       </template>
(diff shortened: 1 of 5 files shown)


chvp (commented):
Between artists and genres seems like a weird place to me. Right before the actions seems like the most sensible place for it (to me).

Also, the track table already uses a lot of its horizontal space and wraps its contents quite easily (which doesn't look that nice, IMO). It might be good to remove the wrapping and ellipsize longer content.

robbevp:
I moved it to right before actions.

I agree that the wrapping is an issue, but that should be for a separate PR.

chvp (approved):

PR #530, opened 2021-07-12
Update issue templates

chvp:
Similar to https://github.com/accentor/api/pull/225/. I also added some text to the pull request asking for screenshots when there are visual changes.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,20 +1,23 @@
-<!--
-Thanks for contributing to Accentor!
-Make sure all GitHub actions (lint & build) will pass and fill out the template.
+<!---
+  Thanks for contributing to Accentor!  Make sure all GitHub actions
+  (lint & build) will pass and fill out the template.
 
-You can tag your PR with the relevant tags and request a review from someone of the team.
-If any changes to your PR are necessary, we will ask for them through the review process.
- -->
+  If any changes to your PR are necessary, we will ask for them
+  throughout the review process.
+  --->
+ 
+<!---
+  Please include a summary of the change and which issue is
+  fixed. Please also include relevant motivation and context. If your
+  pull request includes visual changes (which will probably be the
+  case for anything that isn't a pure logic fix), please include
+  screenshots showing those changes.
+  --->
 
-## Description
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,20 +1,23 @@
...
-If any changes to your PR are necessary, we will ask for them through the review process.
- -->
+  If any changes to your PR are necessary, we will ask for them
+  throughout the review process.
+  --->
+ 
+<!---
+  Please include a summary of the change and which issue is
+  fixed. Please also include relevant motivation and context. If your
+  pull request includes visual changes (which will probably be the
```
robbevp, line 12: Do you have a reference/idea about copyright and album art? It's likely that screenshots will include album art at some point. I'm not sure if and how they fall under copyright laws.
If there are some copyright laws involved, we should mention that the contributor should not upload any copyrighted material and point them to a reference about this

chvp: Under US copyright law, I'm fairly certain this could be considered [fair use](https://en.wikipedia.org/wiki/Fair_use#U.S._fair_use_factors). Both the first and fourth factor seem to imply that the use of copyrighted material would not require permission in this case. Belgian law however does not seem to [have similar exceptions](https://www.ejustice.just.fgov.be/cgi_loi/loi_a1.pl?language=nl&cn=2013022819&caller=list&fromtab=wet#LNK0418). GitHub requires [in their ToS](https://docs.github.com/en/github/site-policy/github-terms-of-service#d-user-generated-content) that a user has the right to post the content they post. I'm not sure from which jurisdiction the copyright law would apply. I would assume from the country of the posting user, but it might also be all the countries of users who viewed the reproduction. I asked some people who are more knowledgeable on this than me, but it might be safer to request that only public domain content is posted.

robbevp: I think it would be good to add a note to encourage images from the public domain ( and maybe also to the bug report template?)

robbevp: It's seems very unlikely that someone would take issue with this usage, but better safe than sorry

robbevp (approved):

PR #532, opened 2021-07-14
Make column widths and alignment in track table a little nicer

chvp:
This aligns the track number to the right (generally recommended for numbers, allows easier comparison) and makes sure the end alignment for the plays column doesn't look too weird (at least it looked a little weird to me).

Before:
screenshot_2021-07-14-134616

After:
screenshot_2021-07-14-134727

### DIFF
--- a/src/components/TracksTable.vue
+++ b/src/components/TracksTable.vue
@@ -125,7 +125,7 @@ export default {
         text: "#",
         value: "number",
         sortable: false,
-        align: "center",
+        align: "end",
         width: "1px",
       },
       {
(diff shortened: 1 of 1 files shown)


robbevp (commented):
`width: 1px` feels very hacky to me. 
Would something like `max-width: min-content` not work? That feels more aligned with what we are trying to do.

chvp:
Tried it, doesn't work. As far as I could find, `width: 1px` is the accepted way of making sure a table column doesn't grow wider than its content.

robbevp (approved):

robbevp:
> Tried it, doesn't work. As far as I could find, `width: 1px` is the accepted way of making sure a table column doesn't grow wider than its content.

ok, let's go with this solution then

PR #537, opened 2021-07-17
Improve grid views on xs screens

robbevp:
We currently show 1 card per row in all grid views on xs screens. This makes navigating a bit unhandy and requires a lot of scrolling (esp in paginated elements like the home page).
I've changed the grid to show 2 cards per row on all iterators.

This caused the actions in the AlbumCard to warp, so I've removed the margin between the actions on xs screens. (They'll still warp on some very small screens, but this is unavoidable). This was only necessary for the album actions, but I've removed this margin everywhere, for more consistency.

Previous views:
Screenshot 2021-07-17 at 10 26 02
Screenshot 2021-07-17 at 10 25 53

New views:
Screenshot 2021-07-17 at 10 25 18
Screenshot 2021-07-17 at 10 24 52
Screenshot 2021-07-17 at 10 24 42

### DIFF
--- a/src/components/AlbumActions.vue
+++ b/src/components/AlbumActions.vue
@@ -7,7 +7,7 @@
             @click.stop.prevent="startTracks"
             :disabled="playableTracks.length === 0"
             color="primary"
-            class="ma-sm-2"
+            class="ml-0 ma-1 ma-sm-2"
             text
             icon
             small
@@ -25,7 +25,7 @@
             @click.stop.prevent="addTracks"
             :disabled="playableTracks.length === 0"
             color="success"
-            class="ma-sm-2"
+            class="ma-1 ma-sm-2"
             text
             icon
             small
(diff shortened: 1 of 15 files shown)


chvp (commented):
Can we reduce the margin instead of removing it entirely? It doesn't look that good to me without it.

robbevp:
> Can we reduce the margin instead of removing it entirely? It doesn't look that good to me without it.
Adding `ma-1` and removing the left/right margin from the first/last items works, but will wrap sooner

chvp:
What's the minimum width we can support in that case?

robbevp:
With some margin: 432px, with no margin: 384px;

chvp:
I'd prefer the margin then. There aren't that many screens still that are less than 432 pixels wide, and the difference isn't that large.

chvp (approved):

PR #544, opened 2021-07-20
Don't wrap track actions

### LINKED ISSUE #541: TrackActions wrap unnecessarily
##### Reproduction steps

  1. Open a few albums/artists.
  2. Notice that the menu icon is below the flag instead of to the right.
  
##### Expected behaviour
The actions should not wrap

##### System information
Chrome 91 on macOS 11.4

chvp:
Due to #532 limiting the width of the last column, the content of those cells would wrap unnecessarily. This makes sure that column doesn't wrap. I still don't really get why I didn't see this when making that PR (you can see in the screenshots there that it wasn't a problem), but :woman_shrugging:.

Before:
screenshot_2021-07-20-140143

After:
screenshot_2021-07-20-140047

Fixes #541.

### DIFF
--- a/src/components/TrackActions.vue
+++ b/src/components/TrackActions.vue
@@ -1,5 +1,5 @@
 <template>
-  <span>
+  <span class="text-no-wrap">
     <VTooltip bottom :disabled="track.length !== null">
       <template v-slot:activator="{ on }">
         <span v-on="on">


robbevp (commented):
I think this is caused by #537 (since that modified the margin on the flag icon).
This works, but I'm not sure if it is the way to go.

Ideally we could present a series of actions both with more and less spacing between the icons. (more spacing being used on an object's page and in card on large screens, less spacing being used in cards on small screens and inside tables).
Maybe we could add a `actions--wide` or `actions--compact` class to the wrapper and modify the margins of each icon based on that?

chvp:
That seems like an orthogonal concern to me. Wrapping the track actions feels like something we should want to avoid regardless of screen size.

robbevp (approved):

PR #545, opened 2021-07-21
Improve spacing between actions

robbevp:
Re #544 #541 
I've reworked the system for actions spacing, so we can modify the space between all icons from the parent component.
this way the `EditReviewComment` icon can have different spacing depending on the context it is in.
I've still added a `mr-0` utility class to the last icon (for moderator and admins), this is mostly too prevent the icons from wrapping sooner than necessary.

This doesn't result in much visual difference from before. The biggest place to notice a difference is the Flag icon in the `TrackActions`, which now has the same spacing as the other elements.

### Before:
Screenshot 2021-07-21 at 14 07 56
Screenshot 2021-07-21 at 14 08 06
Screenshot 2021-07-21 at 14 08 34
Screenshot 2021-07-21 at 14 08 53

### After:
Screenshot 2021-07-21 at 14 11 03
Screenshot 2021-07-21 at 14 10 52
Screenshot 2021-07-21 at 14 10 35
Screenshot 2021-07-21 at 14 10 15

### DIFF
--- a/src/Main.vue
+++ b/src/Main.vue
@@ -33,7 +33,7 @@ a {
 }
 
 @media (min-width: map-get($grid-breakpoints, "sm")) {
-  .sm\:actions--wide {
+  .actions--sm-wide {
     .actions__button {
       margin: 8px 16px 8px 0;
     }
--- a/src/Main.vue
+++ b/src/Main.vue
@@ -15,7 +15,28 @@ export default {
 </script>
 
 <style lang="scss">
+@import "~vuetify/src/styles/styles.sass";
+
 a {
   text-decoration: none;
+}
+
+.actions {
+  .actions__button {
+    margin: 4px 8px 4px 0;
+  }
+}
+.actions--wide {
+  .actions__button {
+    margin: 8px 16px 8px 0;
+  }
+}
+
+@media (min-width: map-get($grid-breakpoints, "sm")) {
+  .sm\:actions--wide {
+    .actions__button {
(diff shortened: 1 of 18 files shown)


chvp (commented):
Why the weird class names? What's wrong with `actions-button` and `wide-actions`? Also, I think the `-sm` postfix is the most commonly-used option for making something only available to larger screens. (At least that's what I'm used to, haven't seen `sm:` prefix before.) Although Vuetify seems to use a `-sm-` infix, so it might be better to go for that.

robbevp:
BEM offers more clarity to what the classes signify. With semantic style classes like `actions-button` and `wide-actions` the relationship between the cla is less clear.

The `sm:` style for breakpoints (popularized by tailwindcss) is IMHO more elegant, but there is a valid argument is using the infix system from vuetify for more consistency

chvp:
Consistency with vuetify seems like the most important thing to me, and they also don't use BEM (at least I don't think so?). It's also not something that is done in the rest of the codebase, so introducing it here feels a little weird to me.

robbevp:
Vuetify uses BEM in combination with utility classes, exactly what I have done here

chvp:
Huh, I don't think I've ever seen that in their documentation. OK then, but then we should probably convert our other custom classes to use the same system (in another pull request though). I would still change the `sm:` prefix to an `-sm-` infix though.

chvp (approved):

Issue #230, opened 2020-07-31
Playlist is affected when track is merged


robbevp:
If a track in the playlist is involved in a merge where its metadata is not being kept, the track is removed from the playlist (since the track is removed from the store). If that is the currently playing track, the audio stops abruptly.

Maybe figure out if we can update the playlist to include the track that it ws merged with? 
(To be honest, this if very low priority for me, but might be good to keep track of this problem)

PR #546, opened 2021-07-21
Add GitHub action to create a release when a tag is pushed

chvp:
Same as https://github.com/accentor/api/pull/230


robbevp (approved):
Not for this PR, but should we start keeping a CHANGELOG.md?
That way this releases will make more sense

robbevp:
Ignore the changelog remark. I just saw the example linked in the api PR and that looks fine

PR #551, opened 2021-07-24
Reset scroll position on navigate

### LINKED ISSUE #549: Scroll position does not reset on navigate
##### Reproduction steps

  1. Go to page that scrolls (fe. artists with a lot of albums or homepage). 
  2. Scroll down a bit
  3. Click on element to navigate
  4. Notice how scroll position is at the same place as before

(This is obvs more noticable on smaller resolution screens or small viewports)

##### Expected behaviour
After navigating from one page to the next, a user should start at the top of the page again.

##### System information
Chome 91 on macOS 11.4

(This might be an issue with vuetify, or might have been there all along and I only just noticed it now)

robbevp:
Fixes #549 

I've chosen to reset the scroll position when we navigate to a different path. This means that the scroll will not reset if when using the pagination or searching

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -186,6 +186,13 @@ const router = new Router({
     },
     { path: "*", redirect: "/app/" },
   ],
+  scrollBehavior(to, from, savedPosition) {
+    if (to.path === from.path) {
+      return savedPosition;
+    } else {
+      return { x: 0, y: 0 };
+    }
+  },
 });
 
 store.watch(


src/router.js
```diff
@@ -186,6 +186,13 @@ const router = new Router({
     },
     { path: "*", redirect: "/app/" },
   ],
+  scrollBehavior(to, from, savedPosition) {
+    if (to.path === from.path) {
```
chvp, line 190: Does this happen?

chvp (approved):

src/router.js
robbevp: Yes, if you use the pagination or search

chvp: Ah, right, `path` doesn't include the query.

PR #555, opened 2021-07-24
Switch api implementation to `@accentor/api-client-js`

robbevp:
Switch the api over to `@accentor/api-client-js`

I tested this for all generic methods (index, CRUD, destroy all, merge) with a few objects and also tested the special cases:
* Start/show rescan
* Create play
* Login/Logout


chvp:
I assume this is blocked on https://github.com/accentor/api-client-js/issues/25?

robbevp:
> I assume this is blocked on [accentor/api-client-js#25](https://github.com/accentor/api-client-js/issues/25)?

I've updated this to set `loaded` when commiting and bumped the api-client version

chvp (approved):

PR #563, opened 2021-08-04
Show image preview in artist and album form

robbevp:
Depends on accentor/api#236

Shows a preview of the current image or uploaded image for artists and albums while editing. This way a mod/admin knows whether an album/artist already has an image and can see the selected image before actually uploading. If no image is present, we show the fallback.
Since this component would only work for images, I've renamed it to `ImagePicker`.

### Design
I didn't find a clear Material Design example for this kind of component, so I relied on my own intuition and the general MD guidelines. Feedback is very welcome.
(I had another idea where the actions would be overlays on the image, but I felt that that would be better suited for a component that lets you upload multiple images.)

Screenshot 2021-08-04 at 09 30 44
Screenshot 2021-08-04 at 09 30 21
Screenshot 2021-08-04 at 09 30 11


src/components/AlbumForm.vue
```diff
@@ -292,6 +296,9 @@ export default {
     ...mapGetters("labels", {
       sortedLabels: "labelsByName",
     }),
+    imagePreviewSrc() {
+      return this.newAlbum.image?.data ?? this.album?.image500;
+    },
```
chvp, line 301: This seems unused?

robbevp: Correct, I forgot to remove it.

chvp (approved):

PR #564, opened 2021-08-06
Allow streaming of transcoded audio

robbevp:
Allows the user to select a prefered codec conversion and stream audio in that conversion.

Confirmed through API log and returned headers that the selected codec conversion was returned in the stream.

The current implementation has one drawback: when a user selects a (different) codec conversion and submits the settings, the currently playing song resets and start again (but in the selected conversion). I don't think this will happen often, so have chosen to leave this as such for now.

Screenshot 2021-08-06 at 10 15 09


chvp (changes requested):
I think we need some sort of handling for when the user has a conversion selected that no longer exists. A getter in the store that handles this case seems like the best option to me. Also, this does not allow non-moderator-users to configure the codec conversions, since the loading of codec conversions was delayed to the opening of the library settings page in #358 (which a non-moderator-user will never be able to open).

chvp (approved):

PR #577, opened 2021-08-13
Fix albums sort order when showing in reverse chronological order

### LINKED ISSUE #574: Sorting of albums doesn't match sorting in android app
##### Reproduction steps

  1. Go to Home screen in app
  2. Go to home screen in web interface
  3. See that sorting of albums with the same release date is inconsistent.
  
I'm not sure if this is an issue with web or with the app. Feel free to move to web if you think the issue is there

##### Expected behaviour
Sorting should be the same

##### System information
v0.6.3

robbevp:
Fixes #574 

This issue arose because we reversed the whole array when we wanted the newest releases first. I've modified our sorting function so that we can sort the releases in both orders and then always sort A-Z


chvp (changes requested):
`compareAlbumsByReleaseFirst` has a few other usages as well, so this will break those.

robbevp:
> `compareAlbumsByReleaseFirst` has a few other usages as well, so this will break those.

I checked (since I suspected that we had to call `sort(compareAlbumsByReleaseFirst())` - but turns out this still works

chvp:
It would still make the code a whole lot clearer if there was an explicit call at the other usages, IMO.

chvp (changes requested):
Needs linting.

chvp (approved):

PR #578, opened 2021-08-14
Fix spacing issue in VDataFooter

robbevp:
A small fix to correctly align the pagination in all VDataFooter's. This wouldn't work if we added extra features to the VDataFooter, but I don't see that happening soon.

If/when this is fixed in vuetify itself, we can simply revert this commit.

## Views with fix
Screenshot 2021-08-14 at 10 21 56
Screenshot 2021-08-14 at 10 23 12

### DIFF
--- a/src/Main.vue
+++ b/src/Main.vue
@@ -46,4 +46,9 @@ a {
 .white-space-nowrap {
   white-space: nowrap;
 }
+
+// Temporary fix until https://github.com/vuetifyjs/vuetify/issues/13678 is resolved
+.v-data-footer {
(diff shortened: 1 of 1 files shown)


chvp (commented):
Is this really a big enough issue to introduce a workaround for? I can't say I had noticed it.

robbevp:
It not a big issue, but it bugs me a lot every time I see it

chvp:
OK, I'll merge after you've fixed the conflicts.

robbevp:
rebased

chvp (approved):

PR #583, opened 2021-08-14
Make sure randomSort does not mutate store

### LINKED ISSUE #582: Sorting randomly should copy array first
##### Reproduction steps

1. Go to home
2. Go to artists
3. See that artists are sorted randomly

##### Expected behaviour

Artists should not be sorted randomly in artists view

robbevp:
Fixes #582

### DIFF
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -218,7 +218,7 @@ export default {
       return items;
     },
     randomSort(items) {
-      const newItems = items;
+      const newItems = [...items];
       for (let i = newItems.length - 1; i > 0; i--) {
         const j = Math.floor(Math.random() * (i + 1));
         [newItems[i], newItems[j]] = [newItems[j], newItems[i]];
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -218,11 +218,12 @@ export default {
       return items;
     },
     randomSort(items) {
-      for (let i = items.length - 1; i > 0; i--) {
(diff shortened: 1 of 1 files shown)


src/views/Home.vue
```diff
@@ -218,11 +218,12 @@ export default {
       return items;
     },
     randomSort(items) {
-      for (let i = items.length - 1; i > 0; i--) {
+      const newItems = items;
```
chvp, line 221: This is not a copy. It's a copy of the reference, so any mutations here will also show up in the original array.

chvp (changes requested):

src/views/Home.vue
robbevp: Weird, since it does work. If I look in the Vue devtools, `this.albums` and `this.randomAlbums` are ordered differently

chvp: Maybe it's a dev mode vs prod mode issue?

chvp:
```
> const a = [1, 2]
undefined
> const b = a;
undefined
> [b[0], b[1]] = [b[1], b[0]]
[ 2, 1 ]
> a
[ 2, 1 ]
```
This is essentially what this code does, as you can see, it doesn't work.

robbevp: I could see the issue before in dev mode. Anyway, I changed the code

chvp (approved):

PR #617, opened 2021-10-03
Runner per location

robbevp:
Re Accentor/api#250 Accentor/api-client-js#34

This PR provides the basic functionality to work with the multiple runners.
It obscures the multiple runners from the user, but presents their attributes together.

To test this locally: 
* Make sure you have the right branch for `api`/`api-client-js`
* Build `api-client-js` locally
* Update `package.json` to use local dependency for `api-client-js`

### DIFF
--- a/src/store/rescan.js
+++ b/src/store/rescan.js
@@ -80,7 +80,7 @@ export default {
         let result = null;
         do {
           result = await api.rescans.show(rootState.auth, id);
-          commit("setRescan", result);
+          commit("setRescan", { id, rescan: result });
           await new Promise((resolve) => setTimeout(resolve, 1000));
         } while (
           rootState.rescan.lastClick > new Date(result.finished_at) ||
@@ -110,7 +110,7 @@ export default {
       try {
         const result = await api.rescans.start(rootState.auth, id);
         result.running = true;
-        commit("setRescan", result);
+        commit("setRescan", { id, rescan: result });
         setTimeout(() => dispatch("show", id), 1000);
         return true;
       } catch (error) {
--- a/src/store/rescan.js
+++ b/src/store/rescan.js
@@ -1,25 +1,77 @@
 import api from "@/api";
+import { fetchAll } from "./actions";
 
 export default {
   namespaced: true,
   state: {
-    rescan: null,
+    rescans: {},
     lastClick: new Date(0),
     loading: false,
   },
   mutations: {
+    setRescans(state, payload) {
+      const oldRescans = state.rescans;
+      state.rescans = {};
+      for (let id in oldRescans) {
+        state.rescans[id] = oldRescans[id];
+      }
+      const loaded = new Date();
+      for (let rescan of payload) {
+        rescan.loaded = loaded;
+        state.rescans[rescan.id] = rescan;
+      }
+    },
+    setRescan(state, { id, rescan }) {
+      const oldRescans = state.rescans;
+      state.rescans = {};
+      for (let id in oldRescans) {
+        state.rescans[id] = oldRescans[id];
+      }
+      rescan.loaded = new Date();
+      state.rescans[id] = rescan;
+    },
+    setStartLoading(state) {
+      state.startLoading = new Date();
+    },
+    removeOld(state) {
+      const oldRescans = state.rescans;
+      state.rescans = {};
+      for (let id in oldRescans) {
+        if (oldRescans[id].loaded > state.startLoading) {
+          state.rescans[id] = oldRescans[id];
+        }
+      }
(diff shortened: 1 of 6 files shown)


src/store/rescan.js
```diff
@@ -1,33 +1,85 @@
...
   actions: {
-    async show({ commit, rootState }) {
+    async index({ commit, rootState, getters }) {
+      if (rootState.rescan.loading) {
+        return true;
+      }
+      try {
+        commit("setLoading", true);
+        do {
+          const generator = api.rescans.index(rootState.auth);
```
chvp, line 60: Did you test this? It crashes here for me because the module is still called `rescan` in `api-client-js` instead of `rescans`.

chvp (changes requested):

src/store/rescan.js
robbevp: I'd made this change to `api-client-js` but hadn't pushed it yet 🤦

chvp (changes requested):
The reload after scans have finished seems broken (at least it didn't trigger for me). Also, it seems a bit weird to me that one of the motivating factors is to allow admins more granularity when rescanning, but then not expose this granularity in the interface. Combining the location and rescan sections seems like the best option to me to fix this.

robbevp:
I updated this to fix the reload and to allow a moderator/admin to scan one specific locations.

I wouldn't combine the rescans settings and the location settings. The rescan settings get used very often, while the locations are rarely changed (and accidentaly removing a location would have disastruous consequences).

I've still kept the idea the you can just see the combined stats/errors/warnings. While I think we'll want to give more details/control in the future, I'd want to start small, to get a good grip of which things we want to see and what can be abstracted away.

src/store/rescan.js
```diff
@@ -42,13 +94,24 @@ export default {
...
+        return false;
+      }
+    },
+    async start({ commit, dispatch, rootState }, id) {
       commit("setLastClick", new Date());
       try {
-        const result = await api.rescan.start(rootState.auth);
+        const result = await api.rescans.start(rootState.auth, id);
         result.running = true;
         commit("setRescan", result);
```
chvp, line 113: This fails here for me. `setRescan` expects something of the shape `{ id, rescan }`, but you just pass `result` here. (Which has an `id` property, but not a `rescan` property.)

src/store/rescan.js
```diff
@@ -1,33 +1,85 @@
...
       if (rootState.rescan.loading) {
         return true;
       }
       try {
         commit("setLoading", true);
         let result = null;
         do {
-          result = await api.rescan.show(rootState.auth);
+          result = await api.rescans.show(rootState.auth, id);
           commit("setRescan", result);
```
chvp, line 83: Will also need changes here, I think.

chvp (changes requested):
Maybe I'm seeing things, but the start scan button looks disabled somehow?

"Start all" text is gone after pressing it and all rescans have finished. The button no longer looks disabled at this point, but the selector to the right still does.

src/views/App.vue
```diff
@@ -188,7 +188,11 @@ export default {
       this.$i18n.locale = this.locale;
     },
     finishedAt(newValue, oldValue) {
-      if (!this.loading && typeof oldValue !== "undefined") {
+      if (
+        !this.loading &&
+        oldValue !== null &&
+        oldValue.getTime() != newValue.getTime()
```
chvp, line 194: Is there a reason not to use `!==` here?

robbevp:
> Maybe I'm seeing things, but the start scan button looks disabled somehow?
> 
> "Start all" text is gone after pressing it and all rescans have finished. The button no longer looks disabled at this point, but the selector to the right still does.

There are some weird bugs going on here. Most of them seem to have their source in VBtnToggle.
For example: the text is still there, but somehow the `color` property is not determined properly.
Screenshot 2021-11-20 at 14 25 44

Since VBtnToggle seemed to make things more difficult (and IMHO doesn't look that great). I just created our own styles for a button-group: (Since we don't actually need any logic on the group, just some styling)

Screenshot 2021-11-20 at 14 47 27
Screenshot 2021-11-20 at 14 47 31

chvp (changes requested):
The dropdown button is very wide, no? Otherwise looks and works fine, thanks!

robbevp:
> The dropdown button is very wide, no? Otherwise looks and works fine, thanks!

Fair point. I modified it a bit, so it now looks like this
Screenshot 2021-12-04 at 13 17 34

chvp (approved):

chvp (approved):

PR #647, opened 2021-11-19
Add recently played albums and artists on home

### LINKED ISSUE #645: Add recently listened albums/artists to homepage
I usually listen to the same album multiple times in short time periods or (re-)discover an artist and want to listen to them a lot. Adding this would make those albums/artists more discoverable. This sort of overlaps with the idea of dynamic playlists/collages, but until we have those it makes sense to add this IMO.

robbevp:
Fixes #645 .

* Reworks `playCountsByTrack` to be `playStatsByTrack` (and track both the count and last played at)
* Adds getters to get plays stats by album and by artist (we only use part of this information now, but will use both *soon*)
* Adds recently played albums and  artists to home

I could see a few different ways to get the playStats for albums and artists, if there are better/more effecient ways to get this info I'm all ears!
I did test this locally with our production api (and NODE_ENV=production) and chrome reported very similar memory usage to the production version we have running

### DIFF
--- a/src/components/TracksTable.vue
+++ b/src/components/TracksTable.vue
@@ -88,7 +88,11 @@
         <TrackGenres :track="props.item" />
       </template>
       <template v-slot:item.play_count="props">
-        {{ playCountsByTrack[props.item.id] || 0 }}
+        {{
+          (playStatsByTrack[props.item.id] &&
+            playStatsByTrack[props.item.id].count) ||
+          0
+        }}
       </template>
       <template v-slot:item.actions="props">
         <TrackActions :track="props.item" />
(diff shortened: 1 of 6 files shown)


src/components/TracksTable.vue
```diff
@@ -88,7 +88,11 @@
         <TrackGenres :track="props.item" />
       </template>
       <template v-slot:item.play_count="props">
-        {{ playCountsByTrack[props.item.id] || 0 }}
+        {{
+          (playStatsByTrack[props.item.id] &&
+            playStatsByTrack[props.item.id].count) ||
+          0
```
chvp, line 94:
```suggestion
          playStatsByTrack[props.item.id]?.count || 0
```
Or does this not work in the templates?

robbevp: This does not work in templates

robbevp:
Removed the albums playcount, since we don't yet know on how to calculate this

chvp (approved):

PR #653, opened 2021-11-20
Allow files to be dropped on ImagePicker

robbevp:
Allow images to be dropped on the current image/placeholder or select button. (This should also allow a user to drag an image from another application/window)

This mirrors the browser's native functionality of dropping files on an `input[type="file"]`:
* If more than one file is dropped, the event is ignored altogether
* If the file does not match the criteria for the file type, the event is ignored


chvp (approved):
This is great, thanks!

PR #658, opened 2021-11-27
Add hidden to track artists

robbevp:
Matches accentor/api#273

* Hides the TrackArtist in the TrackTable if hidden
* Allows editing of this status.

The fields per track artist now look like this:
Screenshot 2021-11-27 at 15 45 14

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "i18n:report": "vue-cli-service i18n:report --src './src/**/*.?(js|vue)' --locales './src/locales/**/*.json'"
   },
   "dependencies": {
-    "@accentor/api-client-js": "^0.15.0",
+    "@accentor/api-client-js": "^0.16.0",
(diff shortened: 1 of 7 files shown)


chvp (changes requested):
I feel like there should be a bit more explanation on what hide in overview means. I don't think it will be obvious to first-time users. Maybe an info icon with a longer explanation as an (immediate) tooltip?

robbevp:
> I feel like there should be a bit more explanation on what hide in overview means. I don't think it will be obvious to first-time users. Maybe an info icon with a longer explanation as an (immediate) tooltip?

Good idea. I've added this. I tried to keep the copy concise and clear, but other suggestions/wordings are very welcome.

chvp (dismissed):

chvp (approved):

PR #660, opened 2021-11-28
Add play stats page

### LINKED ISSUE #659: Allow user to see play stats
### Discussed in https://github.com/accentor/web/discussions/644

<div type='discussions-op-text'>

<sup>Originally posted by **robbevp** November  6, 2021</sup>
Since we track plays - we should also provide the user some stats about their listening habits:

Some first ideas:
Select time periods:
* last 7 days
* last month
* last 12 months
* current calendar year
* previous calendar year
* all time
* custom range

Show stats both for the amount of plays and the amount of time listened (so play count * track length)

Show following stats:
* top x artists
* top x tracks
* top x albums

Some other stats that could be interesting:
* Total time spent listening
* % of library that has been played (in # of tracks)

Other ideas?</div>

robbevp:
Re #659

Done:
* [x] Select for preset of custom date range
* [x] % of library played
* [x] Top 10 tracks
* [x] Compare by play count or track length
 
Some notes:
* I didn't use a library for the visualisations. I feel most libraries either make too much design choices, are huge dependencies and/or are confusing to use/modify. Since we will need only a few graphs, I'm more comfortable generating our own svg's.
* I kept most of the logic outside of the store. Since everything starts from a filtered list of plays, I thought it easier to just do that in the general component.
* I wanted to make make the individual components unaware/independent of their context: ie. the `PercentagePlayedCard` can also be used to give a percentage for an artist/a genre/..., as long as the right `plays` and `tracks` are passed to calculate the percentage

### DIFF
--- a/src/views/Stats.vue
+++ b/src/views/Stats.vue
@@ -5,7 +5,7 @@
         <h1 class="text-h4">{{ $t("common.stats") }}</h1>
       </VCol>
       <VCol class="d-flex justify-end">
-        <VSwitch v-model="useTrackLength" :label="$t('stats.byTrackLength')" class="mr-4" />
+        <VSwitch v-model="useTrackLength" :label="$t('stats.useTrackLength')" class="mr-4" />
         <DateRangeSelect @input="(e) => (period = e)" />
       </VCol>
     </VRow>
--- a/src/views/Stats.vue
+++ b/src/views/Stats.vue
@@ -2,7 +2,7 @@
   <VContainer>
     <VRow>
       <VCol>
-        <h1>{{ $t("common.stats") }}</h1>
+        <h1 class="text-h4">{{ $t("common.stats") }}</h1>
       </VCol>
       <VCol>
         <DateRangeSelect @input="(e) => (period = e)" />
--- a/src/views/Stats.vue
+++ b/src/views/Stats.vue
@@ -4,6 +4,9 @@
       <VCol>
         <h1>{{ $t("common.stats") }}</h1>
       </VCol>
+      <VCol>
+        <DateRangeSelect @input="(e) => (period = e)" />
+      </VCol>
     </VRow>
   </VContainer>
 </template>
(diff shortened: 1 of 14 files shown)


robbevp:
@chvp I think it would be good to have a first review of this, before I invest too much time.
Depending on what you want to do, we could also merge a minimal first version and then add additional cards/graphs in subsequent PRs.

chvp (changes requested):
General idea is definitely good. I'd prefer to merge this without adding more graphs/stats so that we can keep PR sizes small. So, the few things I'd change:
 * The title of the percentage card doesn't overflow correctly (I wouldn't expect words to be split).
 * The "by track length" toggle isn't translated.
 * I would stack the toggle/settings horizontally instead of vertically

chvp:
Ah also, `formatTime` should probably be extended to handle hours. `1370:05` isn't very useful.

robbevp:
> General idea is definitely good. I'd prefer to merge this without adding more graphs/stats so that we can keep PR sizes small. So, the few things I'd change:
> 
>     * The title of the percentage card doesn't overflow correctly (I wouldn't expect words to be split).
> 
>     * The "by track length" toggle isn't translated.
> 
>     * I would stack the toggle/settings horizontally instead of vertically

Good ideas. All implemented

src/views/Stats.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <VContainer>
+    <VRow>
+      <VCol>
+        <h1 class="text-h4">{{ $t("common.stats") }}</h1>
+      </VCol>
+      <VCol class="d-flex justify-end">
```
chvp, line 7: Maybe I'm missing something, but shouldn't this be a `VRow` to make this horizontal?

robbevp: A `VRow` can't be a direct child of itself (or at least, that doesn't look ok).
To use a VRow, it would have to be
```html
<VCol>
  <VRow>
    <VCol>
      <VSwitch />
     </VCol>
     <VCol>
       <DateRangeSelect />
     </VCol>
```
But that felt like a lot of unnecessary structure for to apply two css rules

chvp (approved):

PR #672, opened 2021-12-11
Improve selected date range & useTrackLength in play stats

robbevp:
re #659

Improves the handling of the date range and useTrackLength on the stats by:
* explicitly showing the selected range on the page (especially helpful to clarify what the preset options actually mean)
* Showing the selected range in the pages' title
* storing and retrieving `useTrackLength` in the query.
* storing and retrieving the selected period in the query.
  When interpreting a date from the query, we don't know if it matched a query, so we always opt for `customRange` rather then trying to figure out if this was a preset.
  
## some points of discussion
* I've chosen to create a new history state, rather than to replace the current one when switching periods. That way, a user can easily go back to their previous selection. I'm also fine with using replace
* When using the 'all time' preset, this shows as "01/01/1970" as the start date. Maybe we need to make a special case for this?
* (Not for this PR?): the extracted logic for formatting dates might also be helpful elsewhere (ie. in `store/albums`
* (Not for this PR): It might be handy to group all files with helper functions (filters, reducers, comparators and now dates) in a specific folder.


chvp (changes requested):
I'm not sure about putting the range in the page title. I also feel like we should be able to put the period logic in the datepicker component, where we could special-case on the presets (and therefore not have to revert to custom date when returning to the page).

That being said, I'm also quite ambivalent on the addition of the params to the location. It doesn't feel like it adds that much/would be that useful to me? If you feel otherwise I'm fine with adding it though.

robbevp:
> I'm not sure about putting the range in the page title. 
I've removed this. Didn't feel strongly about it

> I also feel like we should be able to put the period logic in the datepicker component, where we could special-case on the presets (and therefore not have to revert to custom date when returning to the page).
This was a good suggestion. It means giving up two-way binding for this component, but makes the logic a lot simpler.

chvp (approved):

PR #681, opened 2021-12-27
Pause player when pressing spacebar

rien:
This lets the `Player` component listen to key events and pause/unpause the player when the spacebar is pressed. If the playlist is finished (but not empty), pressing spacebar will start playing the playlist from the beginning. It essentially mimics clicking the play/pause button.

It will also prevent the webpage from scrolling down (default browser behavior).

 

- [x] Tested if pausing works 
- [x] Tested if unpausing works
- [x] Tested if there are no errors when the playlist is empty

### DIFF
--- a/src/components/Player.vue
+++ b/src/components/Player.vue
@@ -63,7 +63,7 @@
           <VIcon>mdi-skip-previous</VIcon>
         </VBtn>
         <VBtn
-          @click="setPlaying(!playing)"
+          @click="togglePlaying()"
           :disabled="!playlistTracks.length"
           icon
           class="controls__button"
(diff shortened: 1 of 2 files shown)


src/components/Player.vue
```diff
@@ -171,6 +171,15 @@ export default {
         this.setPlaying(false);
       });
     }
+    window.addEventListener("keydown", (e) => {
+      if (e.code === "Space") {
+        if (this.playlistTracks.length > 0) {
```
robbevp, line 176: Maybe it's a good idea to put this check in `store/player.js`? This way it also applies to all other events that modify this (like the mediaSession controls above).

rien: I've changed `setPlaying` to check the playlist length first before changing the value and added a `togglePlaying` action.

robbevp (approved):

PR #683, opened 2021-12-29
Improve performance of reducers for stats

robbevp:
This should improve the performance of the stats page. 
I changed the way the counts and times are calculated, so that they try to reduce the amount of lookups that have to happen.

@chvp can you test this on your device with your play stats? I generated 40.000 plays locally, but couldn't really reproduce the performance issues (in chrome or firefox). Feel free to take over the PR if you have other ideas that could improve performance

### DIFF
--- a/src/reducers.js
+++ b/src/reducers.js
@@ -11,31 +11,29 @@ export function calcPlayCountForTracks(plays) {
 }
 
 export function calcPlayTimeForTracks(plays, tracks) {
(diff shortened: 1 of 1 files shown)


chvp (approved):
There's still a noticeable delay, but it's definitely a lot faster already. I can't think of any obvious performance improvements right now, so I'll merge this. If we think of any later, we can always create a separate PR.

PR #684, opened 2021-12-29
Add punchcard to stats

robbevp:
Adds a simple punchard to the stats, showing when in the week a user listens to music.
If there are no plays for an hour, nothing in shown. For hours where there is at least on play, the dots range between a radius of 1 and 11 (in the svg grid) - relative the the hour with the highest count.

I've left a few comments in the svg, to make sure we can remember where some of the numbers come from.

Screenshot 2021-12-29 at 12 50 31

### DIFF
--- a/src/components/PlaysPunchcard.vue
+++ b/src/components/PlaysPunchcard.vue
@@ -1,0 +1,124 @@
+<template>
+  <VCard class="pa-2">
+    <VCardTitle> {{ title }}</VCardTitle>
+    <!-- 
+      The viewbox represents:
+      24 * 22 (For the dots) + 25 (for the y-axis) + 11 (for the final x-axis label)
+      by
+      7 * 22 (for the dots) + 10 (for the x-axis)
+     -->
+    <svg viewBox="0 0 569 174" class="punchcard mx-4 mt-1">
+      <!-- Offset 11 (half dot) + 1 (to visually center) -->
+      <g y-axis transform="translate(0,12)">
+        <text
+          class="grey--text text--darken-2 text-caption"
+          fill="currentColor"
+          x="28"
+          :y="2 + 22 * index"
+          text-anchor="end"
+          v-for="(label, index) in yLabels"
+          :key="label"
+        >
+          {{ label }}
+        </text>
+      </g>
+      <!-- Offset 40 (for y axis) by 176 (area for dots) -->
+      <g x-axis transform="translate(30,164)">
+        <text
+          :x="hour * 22"
+          y="5"
+          :text-anchor="
+            index === 0
+              ? 'start'
+              : index === xLabels.length - 1
+              ? 'end'
+              : 'middle'
+          "
+          fill="currentColor"
+          class="grey--text text--darken-2 text-caption"
+          v-for="(hour, index) in xLabels"
+          :key="hour"
+        >
+          {{ hour }}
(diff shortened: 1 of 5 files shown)


chvp (commented):
Is there a reason for having 11 options of widths? We should be able to just calculate the exact required radius, right?

robbevp:
> Is there a reason for having 11 options of widths? We should be able to just calculate the exact required radius, right?

I don't follow. The radius ranges between 1 and 11, but the actual radius is calculated and rounded to 3 points. We could just use the whole number, but you would never see the difference between 1.23456789 and 1.23498765

chvp:
Ah, I misunderstood your initial comment. Any reason for the minimum of 1? Also, what is scaled for the circle? The radius or the area? (For visualisations like this, it should be the area.)

robbevp:
> Ah, I misunderstood your initial comment. Any reason for the minimum of 1? 

I found that adding a kind of baseline made the graphic more clear (especially when the actual visualisation is smaller)

> Also, what is scaled for the circle? The radius or the area? (For visualisations like this, it should be the area.)

The svg scales up and down depending on the size it gets (but with a fixed aspect-ratio). Everything within the svg is placed and sized based on the viewbox, so it should always look the same.

src/components/PlaysPunchcard.vue
```diff
@@ -0,0 +1,124 @@
...
+      return bin;
+    },
+    svgData() {
+      const max = Math.max(...this.binnedPlays);
+      const dots = [];
+      this.binnedPlays.forEach((count, index) => {
+        // We only want to keep/render all the hours where there is at least one play
+        if (count) {
+          // The radius of each dot ranges between 1 and 11
+          const radius = Math.ceil((count / max) * 10000) / 1000 + 1;
```
chvp, line 110: I was talking about this scaling in my comment. You scale the radius linearly, which makes the area scale quadratically. It is the area which should scale linearly.

chvp (changes requested):

src/components/PlaysPunchcard.vue
robbevp: I'm not sure I follow. Can you expand a bit?

chvp: With this code, if in one hour I listened to twice as much tracks as another hour, the radius will be twice as large. This means that the area of the circle will be 4 times as large (since the area of a circle is calculated with (pi*r^2)). So visually, it will look like I listened to 4 times as much music in that hour instead of twice as much.

robbevp: ah - I never got that you meant the are of the circle. So this line should be `Math.ceil(count / max) * 10000) / 500 + 1`?

chvp: No, that just fixes it for the double case. There needs to be a `Math.sqrt` in there somewhere.

chvp: It should be `Math.ceil((Math.sqrt(count) / Math.sqrt(max)) * 10000) / 1000 + 1`

chvp (approved):

chvp:
Can you rebase and fix the conflicts?

robbevp:
> Can you rebase and fix the conflicts?

done

chvp (approved):

PR #685, opened 2021-12-29
Fix TopList readability

robbevp:
Fixes an issue where only part of the count was readable. I ended up placing a second count behind the bar in black. Since we don't know the width of the bar (in pixels) or the length of the count (in pixels), this seemed like the easier solution to me.

To avoid having to change this for three different places, I also extracted the list to a separate `TopList`-component.

Screenshot 2021-12-29 at 15 12 52

### DIFF
--- a/src/components/TopAlbumsList.vue
+++ b/src/components/TopAlbumsList.vue
@@ -12,53 +12,17 @@
         />
       </VCol>
     </VRow>
-    <ol class="top-list">
-      <li v-for="(item, index) in listData" :key="index" class="top-item">
-        <VImg :src="item.image" class="top-item__cover" />
-        <div class="top-item__stat">
-          <span class="top-item__text">
-            <span class="top-item__position font-weight-medium">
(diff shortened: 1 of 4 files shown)


chvp (changes requested):
I'm not a huge fan of the text changing color all of a sudden. Ironically, this doesn't really improve readability for me. I've pushed a patch which gives the following result:
screenshot_2021-12-31-145908
Even though this could still give problems for very long text, I would argue that this wouldn't happen in practice.

robbevp:
If we go for your design, I'd put the label always on the right of the bar, rather than having this determined by the width.

chvp:
Sure, that's also a possibility.

robbevp:
@chvp Thinking about this again. Your solution will be a lot easier than my last suggestion (since then the width of the first item will vary based upon the length of the string behind it. So I'm good with merging this in its current state

chvp (approved):

PR #686, opened 2021-12-30
Allow play stats for artist

robbevp:
Allow the user to see their play stats per artist.

The layout of the stats page stays the same, just with filtered plays/tracks. Plays are currently only filtered based on the track_artists.

To access this page, I added a stats icon to the `ArtistActions`. I'm unsure if this needs to be shown in both the card and on the artist page, both left it like this for now.

Some screenshots showing all visual changes:
Screenshot 2021-12-30 at 10 15 24
Screenshot 2021-12-30 at 10 16 07
Screenshot 2021-12-30 at 10 16 14
Screenshot 2021-12-30 at 10 24 57

### DIFF
--- a/src/locales/nl.json
+++ b/src/locales/nl.json
@@ -374,6 +374,7 @@
 				"minutes": "Minuten | Minuut | Minuten"
 			}
 		},
+		"percentageArtistPlayed": "Percentage van {artist} dat je besluisterd hebt:",
 		"percentageLibraryPlayed": "Percentage van muziekbibliotheek dat je besluisterd hebt:",
 		"topAlbums": "Meest besluisterde albums",
 		"topArtists": "Meest besluisterde artiesten",
--- a/src/locales/nl.json
+++ b/src/locales/nl.json
@@ -374,8 +374,8 @@
 				"minutes": "Minuten | Minuut | Minuten"
 			}
 		},
-		"percentageArtistPlayed": "Percentage van {artist} dat je besluisterd hebt:",
-		"percentageLibraryPlayed": "Percentage van muziekbibliotheek dat je besluisterd hebt:",
+		"percentageArtistPlayed": "Percentage van {artist} dat je beluisterd hebt:",
+		"percentageLibraryPlayed": "Percentage van muziekbibliotheek dat je beluisterd hebt:",
 		"topAlbums": "Meest besluisterde albums",
 		"topArtists": "Meest besluisterde artiesten",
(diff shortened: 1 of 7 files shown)


chvp (changes requested):
I would not show this on the card. It's not that much of a primary action, IMO.

robbevp:
I've updated this to only show the stats button on the artist's page

src/locales/nl.json
```diff
@@ -373,6 +374,7 @@
 				"minutes": "Minuten | Minuut | Minuten"
 			}
 		},
+		"percentageArtistPlayed": "Percentage van {artist} dat je besluisterd hebt:",
 		"percentageLibraryPlayed": "Percentage van muziekbibliotheek dat je besluisterd hebt:",
```
chvp, line 378:
```suggestion
		"percentageArtistPlayed": "Percentage van {artist} dat je beluisterd hebt:",
		"percentageLibraryPlayed": "Percentage van muziekbibliotheek dat je beluisterd hebt:",
```

chvp (changes requested):
I get the following warning in the console:
```
[Vue warn]: Invalid prop: type check failed for prop "extended". Expected Boolean, got String with value "true".

found in

---> <ArtistActions> at src/components/ArtistActions.vue
       <Artist> at src/views/artists/Artist.vue
         <VMain>
           <App> at src/views/App.vue
             <VApp>
               <Main> at src/Main.vue
                 <Root>
```

Also, this is for a separate issue/PR, but we need better content if there are no plays for the selected period. I selected an artist I know I listen a lot to, but haven't listened to in the past week, and thought this was broken until I figured out I had to change the period.

robbevp:
> I get the following warning in the console:
> 
> ```
> [Vue warn]: Invalid prop: type check failed for prop "extended". Expected Boolean, got String with value "true".
> 
> found in
> 
> ---> <ArtistActions> at src/components/ArtistActions.vue
>        <Artist> at src/views/artists/Artist.vue
>          <VMain>
>            <App> at src/views/App.vue
>              <VApp>
>                <Main> at src/Main.vue
>                  <Root>
> ```
> 
> Also, this is for a separate issue/PR, but we need better content if there are no plays for the selected period. I selected an artist I know I listen a lot to, but haven't listened to in the past week, and thought this was broken until I figured out I had to change the period.

Fixed this warning (and create an issue to track the "empty" state of the stats page

chvp (approved):

PR #702, opened 2022-01-15
Improve responsiveness of stats page

robbevp:
A POC to improve the responsiveness of the stats page. This doesn't actually improve the speed/performance of the calculations, but just the perceived speed of the page.

The general idea is to create "async computed values". Since this does have built in support in Vue, we create our own way of this behaviour

@chvp what do you think? Based on this we can do this for every component that does this kind of calculations

### DIFF
--- a/src/components/TopArtistsList.vue
+++ b/src/components/TopArtistsList.vue
@@ -26,21 +26,17 @@ export default {
       default: false,
     },
   },
+  data() {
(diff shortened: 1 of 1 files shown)


chvp (commented):
I don't the subjective speed bonus is worth the added ugliness in the code, personally.

PR #748, opened 2022-03-27
Don't show edition description if empty string

robbevp:
Fixes a few rare occurrences where an album has an edition description, but this is just an empty string. 
Before this fix this would show up as:
Screenshot 2022-03-27 at 13 26 08

(This situation seems to occur when you use the edition description field, but then remove the content again.)


chvp (commented):
Would it make more sense to fix this serverside? Feels like empty strings should just not be allowed into the database.

PR #791, opened 2022-05-29
Use play stats endpoint in stats page

robbevp:
Requires/Matches accentor/api#340 accentor/api-client-js#63

This PR uses the new `plays/stats` endpoint of the api and greatly improves the performance of the stats page.

- I've tried to make the changes in the PR limited to using this new endpoint and removing the client-side calculations, without making any other changes.
- I've not yet added the play stats to the store, or used them in other views. The play stats are the page where we need this improvement the most, and I wanted to keep the scope of this PR limited.

This seemed to work well and improve the response times of the stats pages on my development machine. 
@chvp is there a way/an option to try this out with our real data?

## Possible improvements
* [ ] set the `playStats` value after each x fetches of the endpoint. (We currently wait for all stats to be loaded, which becomes longer and longer the more unique tracks a user has listened to in that period)

### DIFF
--- a/src/components/PercentagePlayedCard.vue
+++ b/src/components/PercentagePlayedCard.vue
@@ -39,7 +39,7 @@ import { gsap } from "gsap";
 export default {
   name: "PercentagePlayedCard",
   props: {
-    plays: {
+    playStats: {
       type: Array,
       required: true,
     },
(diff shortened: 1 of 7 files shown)


chvp (approved):
I don't think there is a way to test this with the real-world data, since we don't really have a staging environment and this requires an api and api-client-js merge and bump.

However the changes look good and seem to work locally, so we can try this out in a new release IMO.

robbevp:
Ok. There might be a few improvements to loading the data that might be relevant before merging - but i probably won't have time for the next 2/3 weeks

PR #820, opened 2022-07-06
Make sure user can't add double item to playlist

### LINKED ISSUE #819: User should get feedback if playlist already contains item
A user currently has no way to known whether an item already belongs to a playlist from the dialog.

##### Reproduction steps
1. Create a playlist.
2. Add an item to that playlist.
3. Open the dialog to add the same item to a playlist again
4. Notice that the playlist that already contains the item is not shown differently in any way

##### Expected behaviour
The user should be able to see/know that one of the options already contains this item.

This is related to accentor/api#351 - but we should also try and prevent this error client side. I'd suggest disabling the playlist options that already contain this item (and maybe an extra text?).

##### System information
N/A

robbevp:
Fixes #819 

This playlist fixes two things around double items in playlists:
* The error returned by the api (see accentor/api#352) is now translated and correctly displayed
* The options that already contain the item are disabled. I've also added a message to explain the disabled state (This message is quite verbose, but I don't mind this)

Since the error should not occur that often anymore, I don't think it's a problem that it is shown in its usual place and not in the dialog.

Screenshot 2022-07-06 at 09 37 12

### DIFF
--- a/src/components/AddToPlaylist.vue
+++ b/src/components/AddToPlaylist.vue
@@ -32,7 +32,13 @@
                 :label="$tc('music.playlists', 1)"
                 return-object
                 v-model="selectedPlaylist"
-              />
+              >
+                <template v-slot:item="props">
+                  <VListItem :disabled="props.item.item_ids.includes(item.id)">
+                    {{ props.item.name }}
+                  </VListItem>
+                </template>
+              </VCombobox>
             </VCol>
           </VRow>
         </VContainer>
--- a/src/components/AddToPlaylist.vue
+++ b/src/components/AddToPlaylist.vue
@@ -88,7 +88,7 @@ export default {
     ...mapGetters("auth", ["currentUser"]),
     ...mapGetters("playlists", ["editablePlaylists"]),
     playlistOptions() {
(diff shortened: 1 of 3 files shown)


src/components/AddToPlaylist.vue
```diff
@@ -79,9 +88,13 @@ export default {
     ...mapGetters("auth", ["currentUser"]),
     ...mapGetters("playlists", ["editablePlaylists"]),
     playlistOptions() {
-      return this.editablePlaylists.filter(
-        (p) => p.playlist_type === this.type
-      );
+      return this.editablePlaylists.reduce((options, p) => {
+        if (p.playlist_type === this.type) {
+          p.disabled = p.item_ids.includes(this.item.id);
```
chvp, line 93: Doesn't this edit the playlist in the store?

robbevp: fixed

chvp: Not actually fixed, `[... ]` performs a shallow copy.

chvp (approved):

PR #822, opened 2022-07-06
Add playlists to album's and artist's page

robbevp:
This PR adds a list of playlists to the page for each album/artist. 
I've split up the translations between the album and artist page, so that it would be easier to differentiate between the two in the future.

Example from album's page
Screenshot 2022-07-06 at 14 26 09

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -279,6 +279,7 @@
 				"label": "Hide in overview?",
 				"explanation": "If an artist is hidden, they will only be shown in detailed lists."
 			},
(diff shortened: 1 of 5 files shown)


robbevp:
Depending on the merging order, we should update this PR or #821, so that the playlists in the album/artist's page are also sorted by name

robbevp:
Rebased after #821 was merged and used the sorted playlist on the album's and artist's page

chvp (approved):

PR #940, opened 2023-03-03
Retry playing audio tracks when loading fails

### LINKED ISSUE #909: Retry playing audio file when failure occurs
I've been noticing more issues with audio files not loading - this usually only occurs on the first attempt, so might be caused by some delay if we have to do content-length calculations.

We currently skip the track, so the user doesn't get disturbed or stuck, but this can become annoying in some situations (say, a live album).
We should try loading a track 2 or 3 times if it seems like a slow connection or slow response is at fault.

robbevp:
Fix #909

With these changes, the player will try to load an audio track three times before giving up and going to the next track.

I wasn't yet able to try this in a real life scenario (where such an error might occur). I'll try and use this branch over the next few days, to see if everything works as expected. We might want to wait for these tests before merging.

### DIFF
--- a/src/components/Player.vue
+++ b/src/components/Player.vue
@@ -152,6 +152,7 @@ export default {
       volume: 100,
       muted: false,
       intervalId: 0,
+      tries: 0,
     };
   },
   created() {
@@ -183,6 +184,7 @@ export default {
   },
   watch: {
     async currentTrackURL() {
+      this.tries = 0;
       if (this.currentTrackURL === null) {
         return;
       }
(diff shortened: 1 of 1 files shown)


chvp (commented):
Would it make sense to try to delay the retry? (If it's not too complex.) Network issues are usually there for a longer period of time, so immediately retrying probably won't help?

robbevp:
> Would it make sense to try to delay the retry? (If it's not too complex.) Network issues are usually there for a longer period of time, so immediately retrying probably won't help?

I've added a simple linear delay. 

I've been using the branch (locally build) for the last few days, and have not noticed/experienced any issues, so this should be good to go

chvp (approved):

PR #1066, opened 2023-11-25
Automatically load node modules in devshell

robbevp:
This PR makes sure that loading the devshell (using direnv or nix develop) automatically includes the node modules.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,39 +21,43 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs { inherit system; overlays = [ devshell.overlays.default ]; };
+        web-package = pkgs.mkYarnPackage {
+          pname = "accentor-web";
+          version = "0.32.0";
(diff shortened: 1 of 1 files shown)


chvp (commented):
How does one add a dependency or manually change it's version with this in place?

robbevp:
Ugh, forgot about that.
We could create a script (like we have for the api) to update both the yarn.lock and yarn.nix file.
I would need to be something like
```
yarn install --modules-location $PRJ_DATA_DIR/tmp_yarn
yarn2nix > yarn.nix
rm -rf $PRJ_DATA_DIR/tmp_yarn
```
We can't install to node_modules, since this would try to write to the nix store, but yarn v1 requires us to install the modules in order to update the lockfile (this isn't the case for the most recent version of npm or yarn).

There _might_ also be a way to make this work by appending/prepending the `NODE_PATH` env var, but I haven't been able to try that.

PR #1094, opened 2024-01-16
Provide a download button for track

robbevp:
After the changes in [the api](https://github.com/accentor/api/pull/487), we can now provide a download button for the original audio.
This PR adds that button for moderators and admins (so it can be shown together with the audio file info)

### DIFF
--- a/src/components/TrackActions.vue
+++ b/src/components/TrackActions.vue
@@ -115,6 +115,19 @@
             </VListItem>
           </VList>
         </VMenu>
+        <VListItem
+          v-if="track.length"
+          :href="downloadURL"
+          download
+          target="_blank"
+        >
+          <VListItemIcon>
+            <VIcon color="info">mdi-download</VIcon>
+          </VListItemIcon>
+          <VListItemContent>
+            <VListItemTitle>{{ $t("music.track.download") }}</VListItemTitle>
+          </VListItemContent>
+        </VListItem>
         <VTooltip bottom :disabled="!waitingForReload">
           <template v-slot:activator="{ on }">
             <VListItem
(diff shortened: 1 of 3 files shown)


robbevp:
Example of track actions when the audio file is present:
Screenshot 2024-01-16 at 22 08 31

chvp:
Is there a reason to only add this for moderator+? The API doesn't have this restriction I think?

robbevp:
> Is there a reason to only add this for moderator+? The API doesn't have this restriction I think?

I didn't think about this too much :) It made most sense to put this together with the audio file's info, which is also restricted to mod+. I'd mostly want to avoid having this button in one of two places (depending on your user level).

chvp:
There would be three secondary actions for normal users as well, I think it would make sense to have an overflow menu for every user now. But we can postpone this to a different PR if that's better for you.

robbevp:
> There would be three secondary actions for normal users as well, I think it would make sense to have an overflow menu for every user now. But we can postpone this to a different PR if that's better for you.

Which action(s) for users would you put also in the overflow menu? The flag / review comment?

chvp:
That, and adding to a playlist.

robbevp:
I've updated this so that the download item is available to all users.

> That, and adding to a playlist.
This doesn't feel logical to me. I'd prefer to keep this within direct access.

Either way: Adding the review comment (and/or playlist) action to the overflow menu would also create substantial changes in those components (since they would need to work as standalone buttons, and in the overflow menu). That didn't feel like something that belongs in this PR, so I've left that for now.

chvp (approved):

PR #1115, opened 2024-02-25
Use endpoint for playlist items when adding a new item

robbevp:
This PR depends on accentor/api-client-js#371 and accentor/api#545

Similar to #1114 , this PR tries to address the time it takes to update playlists. The most common interaction (at least, for me) is to add a single item. This gets slower the larger the playlist.


robbevp:
Updated to match new api

chvp (approved):

PR #1167, opened 2024-08-31
Add `application` to auth token params

robbevp:
This uses the new param from https://github.com/accentor/api/pull/622

I went with the format `WEB vX.Y.Z` but am open to other suggestions.

Manual tests:
* Logging in creates an auth token with the appropriate `application` attribute
* Running `yarn build` creates a bundle that directly contains the version from `package.json`

### DIFF
--- a/src/version.js
+++ b/src/version.js
@@ -1,1 +1,0 @@
-export const APPLICATION_VERSION = `WEB v${require("../package.json").version}`;
--- a/src/version.js
+++ b/src/version.js
@@ -1,0 +1,1 @@
+export const APPLICATION_VERSION = `WEB v${require("../package.json").version}`;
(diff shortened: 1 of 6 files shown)


src/version.js
```diff
@@ -0,0 +1 @@
+export const APPLICATION_VERSION = `WEB v${require("../package.json").version}`;
```
chvp, line 1: I would do this through the vue config (e.g. something like this: https://medium.com/hceverything/how-to-show-your-app-version-from-package-json-in-your-vue-application-11e882b97d8c). Reading the package.json in actual application code doesn't feel that nice.

Also, I would just write `Accentor Web v${version}`.

chvp (commented):
I also just published new version of `api` and `api-client-js`. While this doesn't strictly need any of the changes in `api-client-js` it would be better to also bump it in this PR, IMO.

chvp (approved):

PR #1222, opened 2025-03-08
Fix auth param when loading stats

robbevp:
In #1215 we updated the authorization logic to the new tokens. I forgot to switch the argument when loading the play stats (Resulting in an authorization header containing `Bearer [object Object]`. 

I double checked, but could not find any other cases where I forgot to switch the parameter. This is the only place outside of the stores were we access `state.auth` or interact with the api.

I manually tested the loading the stats now work

### DIFF
--- a/src/views/Stats.vue
+++ b/src/views/Stats.vue
@@ -156,7 +156,10 @@ export default {
       await this.$store.dispatch("plays/index");
     },
     async loadPlayStats() {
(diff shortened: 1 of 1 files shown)


chvp:
This is a duplicate of #1221 😉

robbevp:
Lol, somehow missed the notification for that PR

PR #1398, opened 2026-04-28
Stop using removed yarn2nix infrastructure

chvp:
This PR updates nixpkgs, while switching out the build of the nix package with the recommended (not removed) infrastructure for doing so.


chvp:
(Don't merge after approval, I temporarily switched my nixos-config to follow this branch.)

.github/workflows/dependencies.yml
```diff
@@ -19,8 +19,8 @@ jobs:
         with:
           github_access_token: ${{ secrets.GITHUB_TOKEN }}
           nix_path: nixpkgs=https://github.com/NixOS/nixpkgs/archive/refs/heads/nixos-unstable.tar.gz
-      - name: Run yarn2nix
-        run: nix-shell -p yarn2nix --run "yarn2nix > yarn.nix"
+      - name: Run prefetch-yarn-deps
+        run: nix-shell -p prefetch-yarn-deps --run "nix-hash --type sha256 --to-sri $(prefetch-yarn-deps 2>/dev/null)" > yarn.lock.hash
```
robbevp: There might be cases where we'll want to run this locally. Maybe our devshell should contain a command like `hash-yarn-lock` that executes this (and then we can do something like `nix develop .# --command hash-yarn-lock`)?

robbevp (approved):
I added one suggestion, but this is good to merge

I think we might want to consider moving away from yarn classic in some future (but I haven't spent the time looking at the other options and how well they work with nix recently)

.github/workflows/dependencies.yml
chvp: Good point, done.

